PR #28, opened 2024-09-11
initial API compare tool

RubenSandwich:


scripts/api-compare/api-compare.mjs
```diff
@@ -0,0 +1,350 @@
...
+	)
+	.option(
+		'-d, --diff <type>',
+		'Diff type for API type "content": "metadata", "markdown", or "everything"',
+		'markdown'
+	)
+	.option('-r, --drop-keys <keys>', 'Result keys to drop', (value) =>
+		value.split(',').map((key) => key.trim())
+	)
+	.option('-t, --num-of-tests <number>', 'Number of tests', parseInt, 10)
```
RubenSandwich, line 40: let's remove this and just test every route in a version and product that they give us.

scripts/api-compare/api-compare.mjs
```diff
@@ -0,0 +1,350 @@
...
+	.option('-s, --save-output', 'Save output', false)
+
+program.parse(process.argv)
+const options = program.opts()
+
+const mainContentDirectory = './content'
+
+const contentDirMap = {
```
RubenSandwich, line 48: we don't need this as it should now be in `app/utils/productConfig.mjs`

scripts/api-compare/api-compare.mjs
```diff
@@ -0,0 +1,350 @@
...
+	.option('-n, --new-api-url <url>', 'New API URL')
+	.option('-o, --old-api-url <url>', 'Old API URL')
+	.option('-v, --version <version>', 'Version')
+	.option('-p, --product <product>', 'Product')
+	.option(
+		'-a, --api <api>',
+		'API type: "content", "nav-data", "version-metadata", or "content-versions"',
+		'content'
+	)
+	.option(
```
RubenSandwich, line 32: This only makes sense for the `/docs` api route, so leave it out for now.

scripts/api-compare/api-compare.mjs
```diff
@@ -0,0 +1,350 @@
...
+		'-d, --diff <type>',
+		'Diff type for API type "content": "metadata", "markdown", or "everything"',
+		'markdown'
+	)
+	.option('-r, --drop-keys <keys>', 'Result keys to drop', (value) =>
```
RubenSandwich, line 37: `-r` only makes sense for the `/docs` api route, so leave it out for now.

scripts/api-compare/api-compare.mjs
```diff
@@ -0,0 +1,350 @@
...
+			}
+
+			const apiPaths = getAllContentApiPaths(contentPath)
+
+			const randomIndexes = []
+			const numOfRandomIndexes =
+				Math.min(options.numOfTests, apiPaths.length) || 1
+
+			// let apiPathsLeft = [...apiPaths];
+			while (randomIndexes.length < numOfRandomIndexes) {
```
RubenSandwich, line 216: This is the "fuzzy" part where it randomly tests `/doc` routes, I don't think this is necessary anymore.

PR #183, opened 2025-03-12
feat: add isr github action

LeahMarieBush:
### Description

[Asana task 🎟️ ](https://app.asana.com/1/90955849329269/project/1209399665311208/task/1209573328575681)

This PR adds a github action to revalidate the paths of the changed files every time new content is pushed to main. The action gathers the changed files, maps them to the correct urls using the docsPathsAllVersions.json file, and then sends a post request to the revalidate/paths endpoint to refresh the pages for those paths

### Testing

Since this action runs on main when content is changed, there is not a great way to test it in a PR. I do have a test run [here](https://github.com/hashicorp/web-unified-docs/actions/runs/13821176288) that shows it is able to successfully gather the paths, map them, and send a request to revalidate/paths

### DIFF
--- a/.github/workflows/reload-dev-portal.yml
+++ b/.github/workflows/reload-dev-portal.yml
@@ -11,8 +11,6 @@ on:
         description: 'Head SHA for changed files comparison'
         required: true
         type: string
-  push:
-    branches: ['leah/feat/isr-action']
 
 jobs:
   re-load-dev-portal:
@@ -29,7 +27,7 @@ jobs:
       - name: Reload dev portal
         uses: ./.github/actions/reload-dev-portal
         with:
-          base_sha: ${{ inputs.base_sha || github.event.pull_request.base.sha }}
-          head_sha: ${{ inputs.head_sha || github.event.pull_request.head.sha}}
+          base_sha: ${{ inputs.base_sha }}
+          head_sha: ${{ inputs.head_sha }}
           dev-portal-deploy-hook-prod: ${{ secrets.DEV_PORTAL_DEPLOY_HOOK_PROD }}
           revalidate-token: ${{ secrets.REVALIDATE_TOKEN }}
--- a/content/terraform-plugin-mux/v0.18.x/docs/plugin/mux/index.mdx
+++ b/content/terraform-plugin-mux/v0.18.x/docs/plugin/mux/index.mdx
@@ -4,10 +4,6 @@ description: >-
   Use terraform-plugin-mux to combine and translate providers. It minimizes code changes by wrapping existing provider servers.
 ---
 
-
-test
-
-
 # Combining and Translating Providers
 
 The [terraform-plugin-mux](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-mux) Go module is a collection of Go packages for combining (multiplexing) and translating provider servers. It helps minimize provider code changes by wrapping existing provider servers. This functionality is based on the [Terraform Plugin Protocol](/terraform/plugin/how-terraform-works#terraform-plugin-protocol) and [`terraform-plugin-go`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go) provider servers.


package.json
```diff
@@ -29,6 +29,7 @@
 		"react-dom": "^18.2.0",
 		"remark": "^12.0.1",
 		"remark-mdx": "^1.6.22",
+		"semver": "^7.7.1",
```
RubenSandwich, line 32: Is semver being used? I don't see it anywhere.

scripts/map-files-to-urls.mjs
```diff
@@ -0,0 +1,24 @@
...
+const filePaths = process.argv.slice(2)
+
+// Run the main script
+main(filePaths)
+
+function main(filePaths) {
+	const result = filePaths.map((filePath) => {
+		try {
+			const url = getUrlFromFilePath(filePath)
+			return url
```
RubenSandwich, line 17:
```suggestion
			return getUrlFromFilePath(filePath)
```

.github/workflows/dev-portal-prod.yml
RubenSandwich, line 1: Let's remove [the deploy-dev-portal job](https://github.com/hashicorp/web-unified-docs/blob/59da9e773ff90ea2c04b53ed375282b6d9686aed/.github/workflows/dev-portal-prod.yml#L46), as it's no longer needed if we are doing ISR.

.github/workflows/dev-portal-prod.yml
```diff
@@ -49,3 +49,23 @@ jobs:
     steps:
       - name: deploy dev portal production
         run: curl -X POST ${{ secrets.DEV_PORTAL_DEPLOY_HOOK_PROD }}
+
+  trigger-isr:
```
RubenSandwich, line 53: I thought about this more and instead of this being a it's own separate workflow. I think we should make it [into a composite action.](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action#creating-a-composite-action-within-the-same-repository)

Why?
1. Composite actions live in the same "context" as other steps, so a composite action can reuse files/build output from other steps in a job.
  1. This means that we wouldn't need to checkout, npm i, prebuild. We could use all of that from the deploy-udr step.
  
2. I can see us wanting to keep a version of on-demand-isr as a workflow dispatch. (Only this one would need to checkout, npm i, prebuild, etc.) And having a componsite action means we can reuse all of the code in the current jobs of `get-changed-files` and `map-files-to-urls`.

scripts/utils/file-path/url/index.test.mjs
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import { describe, it, expect, vi } from 'vitest'
+import { getUrlFromFilePath } from './index.mjs'
+import { getProductDirectoryFromFilePath } from '../product-directory/index.mjs'
+import { getVersionFromFilePath } from '../version/index.mjs'
+import { PRODUCT_CONFIG } from '../../../../__fixtures__/productConfig.mjs'
+import allDocsPathsJsonMock from '../../../../__fixtures__/docsPathsAllVersionsMock.json'
+
+vi.mock('../product-directory/index.mjs', () => {
```
RubenSandwich, line 13: Good job adding tests. ❤️ 

Can you test that these tests run when prebuild isn't run? So when:
`app/api/versionMetadata.json` & `app/api/docsPathsAllVersions.json` are missing. This will tell us if you are mocking everything correctly.

RubenSandwich (changes requested):
[I think we should move to using a composite actions.](https://github.com/hashicorp/web-unified-docs/pull/183/files#r1992480109) I know you need to roll off this project soon. So I put some pair programming time on your calendar tomorrow for us to talk this through and quickly get it done.

package.json
LeahMarieBush: Semver is used in `getVersionFromFilePath` which `getUrlFromFilePath` uses and it wasn't in the package.json before

RubenSandwich (approved):
🚢

PR #223, opened 2025-03-20
Fix exit code bug in link checker

RubenSandwich:
Fixes the strange exit code we were getting from the `Broken Link Checker (full)` workflow.

https://github.com/hashicorp/web-unified-docs/actions/runs/13951360829/job/39051469623#step:6:7

### DIFF
--- a/.github/workflows/broken-link-check-full.yml
+++ b/.github/workflows/broken-link-check-full.yml
@@ -46,7 +46,7 @@ jobs:
         if: steps.lychee.outputs.exit_code == 0
         run: |
             echo "🎉 No broken links found! 🎉" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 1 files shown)


.github/workflows/broken-link-check-full.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         if: steps.lychee.outputs.exit_code == 0
         run: |
             echo "🎉 No broken links found! 🎉" >> $GITHUB_STEP_SUMMARY
-            - name: Create GitHub Issue From lychee output file
```
RubenSandwich, line 49: This was the line that I believe was causing the strange exit error as it's not a proper shell command. I think I left it in when I refactored this workflow in another PR. 🤦🏻

heatlikeheatwave (approved):

PR #230, opened 2025-03-26
Enable vercel deployment protection bypass

rmainwork:
## Description

Port of https://github.com/hashicorp/dev-portal/pull/2718

## 🗒️ What

Enable vercel authentication bypass for broken link checker tests.

## 🤷 Why

Vercel deployment protection is currently disabled for previews because enabling it would cause the broken link checker to fail. The fix for this is to enable automation bypass which allows an additional token (obtained via github secret) to be passed in the request header allowing programmatic access to preview deployments

## 💭 Anything Else ?

When this PR is merged, some additional actions are needed to finish setting up Vercel auth:

1. Go into Vercel and [enable deployment protection](https://vercel.com/hashicorp/web-unified-docs/settings/deployment-protection).
2. Be sure to select "Only pre-production deployments" from the dropdown

### DIFF
--- a/.github/workflows/build-pr-preview.yml
+++ b/.github/workflows/build-pr-preview.yml
@@ -24,7 +24,8 @@ jobs:
     name: Get Changed Files
     runs-on: ubuntu-latest
     outputs:
-      changed_content_files: ${{ steps.changed-files.outputs.content_all_changed_files }}
+      changed_markdown_content_files: ${{ steps.changed-files.outputs.markdown_content_all_changed_files }}
+
       changed_content_files_count: ${{ steps.changed-files.outputs.content_all_changed_files_count }}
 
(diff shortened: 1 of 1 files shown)


RubenSandwich:
@rmainwork 

1. Can you fix the merge conflict
2. [Here is another place where we are using the link check on dev-portal.](https://github.com/hashicorp/web-unified-docs/blob/main/.github/workflows/broken-link-check-full.yml#L18) Will we require the by-pass header there as well, or is it only on dev-portal previews?

rmainwork:
1. Yep. Fixed the conflict in 69240d1
2. Nope, that one should be okay because it's connecting to prod devdot (which doesn't have Vercel auth enabled)

RubenSandwich (approved):
🚢

PR #235, opened 2025-03-27
Fix content loading

rmainwork:
Add Vercel deployment protection bypass header to both the CI(broken link checker) and the two `fetch()` requests used to fetch internal docs content. The latter was causing HTTP requests to fail before due to the Vercel authentication page preventing access to the Vercel "CDN"(public folder) used to host and share content.

Iterates on #230

### DIFF
--- a/app/utils/file.ts
+++ b/app/utils/file.ts
@@ -1,5 +1,3 @@
-'use server'
-
 /**
  * Copyright (c) HashiCorp, Inc.
  * SPDX-License-Identifier: BUSL-1.1


.github/workflows/build-pr-preview.yml
```diff
@@ -2,14 +2,7 @@ name: Build Preview
...
 on:
-  pull_request_target:
-    types: [opened, synchronize]
-    branches:
-      - main
-      - develop
-  # paths:
-    # Hello Security 👋, we are checking to make sure forked repo PR changed paths are only in content/** inside the job security-check.
-    # We are doing this so we can also reuse this workflow for internal PRs, as pull_request_target also triggers on internal PRs. (As does pull_request)
+  push:
```
RubenSandwich, line 5: @rmainwork because this is file is a `pull_request_target` on main, your changes in this PR will not go into effect in this PR. (This because security is bigger concern with `pull_request_target` actions.) Please duplicate this workflow file into a new push/pull_request workflow file so that you can see your test runs.

rmainwork: Good point! I copied the workflow into another file, pushed it and (with lots of rebasing afterward) tested it to make sure it works - it seems like it does.

.github/workflows/build-pr-preview.yml
```diff
@@ -219,7 +219,8 @@ jobs:
           cd unified-docs-frontend-preview
           echo "HASHI_ENV=unified-docs-sandbox" >> .env
           echo "UNIFIED_DOCS_API=${{ needs.deploy-unified-docs-api-preview.outputs.preview_url }}" >> .env
-
+          echo "VERCEL_AUTOMATION_BYPASS_SECRET=${{ secrets.VERCEL_AUTOMATION_BYPASS_SECRET }}" >> .env
```
heatlikeheatwave, line 222: can you add this env var to the project settings for the frontend-preview too?

rmainwork: Done.

heatlikeheatwave (approved):

app/utils/file.ts
```diff
@@ -12,6 +12,12 @@ const SELF_URL = process.env.VERCEL_URL
 	? `https://${process.env.VERCEL_URL}`
 	: `http://localhost:${process.env.UNIFIED_DOCS_PORT}`
 
+const headers = process.env.VERCEL_URL
+	? new Headers({
+			'x-vercel-protection-bypass': process.env.VERCEL_AUTOMATION_BYPASS_SECRET,
+		})
+	: new Headers()
```
heatlikeheatwave, line 19: Could you add `'use server'` to the top of the file so this is only run server-side?

heatlikeheatwave:
Just left a suggestion, but otherwise lgtm

heatlikeheatwave:
Sorry about closing this! It was a mistake made while refactoring the the repo-sync workflow.

app/utils/file.ts
rmainwork: Added that in 4f2ecae

PR #237, opened 2025-03-28
Mock imports in app/utils/allDocsPaths.test.ts to run without prebuild step

heatlikeheatwave:
[Asana task](https://app.asana.com/0/1209399665311208/1209829823452614/f)

## Changes
Mock imports for `getDocsPaths` for the test suite `app/utils/allDocsPaths.test.ts` so it can be run without prebuild step

## Testing
1. checkout this branch
```
git checkout heat/chore/all-docs-paths-test-sans-prebuild
```

2. If they exist, delete `api/versionMetadata.json`, `api/docsPathsAllVersions.json` and `public/content*`

3. Run the test
```
npm run test app/utils/allDocsPaths.test.ts
```

4. This should be the output:

```
> test
> vitest app/utils/allDocsPaths.test.ts

 ✓ app/utils/allDocsPaths.test.ts (4)
   ✓ getDocsPaths should return an error for an empty productSlugs array
   ✓ getDocsPaths should return an error if there are no paths for an empty productSlugs array
   ✓ getDocsPaths should return filtered docs paths when a non-empty productSlugs array is provided
   ✓ getDocsPaths should return an error if there are no paths for a non-empty productSlugs array

 Test Files  1 passed (1)
      Tests  4 passed (4)
```

### DIFF
--- a/app/utils/allDocsPaths.test.ts
+++ b/app/utils/allDocsPaths.test.ts
@@ -17,6 +17,18 @@ vi.mock(import('@utils/contentVersions'), async (importOriginal: any) => {
 	}
 })
 
+vi.mock('@api/versionMetadata.json', () => {
+	return {
+		default: {},
+	}
+})
+
+vi.mock('@api/docsPathsAllVersions.json', () => {
+	return {
+		default: {},
+	}
+})
+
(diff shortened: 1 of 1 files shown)


RubenSandwich (approved):
🚢

PR #253, opened 2025-04-01
Remove paths so Test & Lint action runs on every pr

heatlikeheatwave:
[Asana task](https://app.asana.com/0/1209399665311208/1209864842550501/f)

This action needs to run on every PR, so the paths restriction needs to be removed from the workflow otherwise it will block non-js file changes


RubenSandwich (approved):
🚢

PR #263, opened 2025-04-02
Remove lock logic from repo-sync workflow

heatlikeheatwave:
[Asana task](https://app.asana.com/0/1209399665311208/1209871197985443/f)

## Changes
- Remove logic that locks a PR as it prevents the sticky comment step in deploy previews from working

### DIFF
--- a/.github/workflows/repo-sync.yml
+++ b/.github/workflows/repo-sync.yml
@@ -110,19 +110,6 @@ jobs:
               throw err
             }
 
-            console.log('Locking conversations to prevent spam')
-            try {
-              await github.rest.issues.lock({
-                ...context.repo,
-                issue_number: pull_number,
-                lock_reason: 'spam'
-              })
-              console.log('Locked the pull request to prevent spam')
-            } catch (error) {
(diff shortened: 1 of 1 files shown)


.github/workflows/repo-sync.yml
```diff
@@ -110,29 +110,6 @@ jobs:
...
-                issue_number: pull_number,
-                lock_reason: 'spam'
-              })
-              console.log('Locked the pull request to prevent spam')
-            } catch (error) {
-              console.error('Failed to lock the pull request.', error)
-              // Don't fail the workflow
-            }
-
-            console.log('Counting files changed')
```
RubenSandwich, line 126: @heatlikeheatwave wouldn't we want to keep this part in? As we likely don't want to me making PRs if no files have changed.

RubenSandwich (changes requested):
I think we likely want to not create PRs if no files have changed.

.github/workflows/repo-sync.yml
heatlikeheatwave: Added those back in. That was a mistake on my part

PR #342, opened 2025-04-15
Remove terraform docs common

rmainwork:
Remove `terraform-docs-common` from `develop` branch of UDR in preparation for the launch of  `terraform-cdk` & `terraform-docs-agents` tomorrow.

---
  - https://app.asana.com/0/0/1209988875612452

### DIFF
--- a/app/api/content/[productSlug]/redirects/route.ts
+++ b/app/api/content/[productSlug]/redirects/route.ts
@@ -18,6 +18,7 @@ const contentDirMap: Record<string, string> = {
 	terraform: 'docs',
 	'terraform-cdk': 'docs',
 	'terraform-docs-agents': 'docs',
+	'terraform-docs-common': 'docs',
 	'terraform-plugin-framework': 'docs',
 	'terraform-plugin-log': 'docs',
 	'terraform-plugin-mux': 'docs',


app/api/content/[productSlug]/redirects/route.ts
```diff
@@ -18,7 +18,6 @@ const contentDirMap: Record<string, string> = {
 	terraform: 'docs',
 	'terraform-cdk': 'docs',
 	'terraform-docs-agents': 'docs',
-	'terraform-docs-common': 'docs',
```
Comment (LeahMarieBush) [inline app/api/content/[productSlug]/redirects/route.ts:21]: Should this be removed? Other products that are not in UDR are in this map

LeahMarieBush (commented):
One small comment, the rest looks good!

app/api/content/[productSlug]/redirects/route.ts
Comment (rmainwork) [reply app/api/content/[productSlug]/redirects/route.ts:21]: Interesting point, I'm not sure. I'm going to talk to Heat later about the weird test failures as well so I might pick their brains about this too..

Comment (heatlikeheatwave) [reply app/api/content/[productSlug]/redirects/route.ts:21]: I don't think this needs to be removed. It's just a mapping of what the content directories are called for each repo

Comment (rmainwork) [reply app/api/content/[productSlug]/redirects/route.ts]: Ended up reverting that in [ff4ffac](https://github.com/hashicorp/web-unified-docs/pull/342/commits/ff4fface5e203cd0ddd62d4012431e87755c7b65)

LeahMarieBush (approved):

PR #345, opened 2025-04-15
Update CODEOWNERS file for terraform-docs-agents

rmainwork:
Tthere's some difficulty locating the correct codeowners for Terraform CDK, so for now we're just launching with `terraform-docs-agents` and leaving `terraform-cdk` for later.

---
  - https://app.asana.com/0/0/1209988968519347


RubenSandwich (approved):
🚢

PR #364, opened 2025-04-29
fix: find doc versions returning too many versions

LeahMarieBush:
### Description

[🎟️  Asana task](https://app.asana.com/1/90955849329269/project/1209399665311208/task/1210100692227122?focus=true)

This PR fixes findDocVersions. Currently it just checks if the path is somewhere in the string of version data. While that works for most cases, there is an edge case where the path could be contained in a larger path but the doc does not exist for that version. Ex. `concepts/providers` would return true for `concepts/providers-and-resources` even though they are different pages. I noticed this issue when testing the terraform-cdk docs. It would return invalid versions since they are included in the version data string but don't exist for that version. I've changed it to check if the path in the version data ends with the fullpath passed into the function instead.

### Testing

To test that it works with docs we currently have
1. Go to [preview](https://unified-docs-frontend-preview-5l5tcz272-hashicorp.vercel.app/terraform/plugin/framework/ephemeral-resources)
2. Verify that the versions returned are the same as they are in prod

To test that it works with terraform-cdk docs
1. Checkout this branch
2. Un-comment terraform-cdk from productConfig.mjs
3. Run `node ./scripts/migrate-content/migrate-content.mjs terraform-cdk -force`
4. Run `npm run prebuild`
5. Start up UDR
6. Go to http://localhost:3000/terraform/cdktf/concepts/providers
7. Verify that the versions in the dropdown on this page match the versions in prod


rmainwork (changes requested):
This looks great. However, I would maybe suggest that we might want to add a test for this? I'm aware that there currently are no tests for this function, so perhaps just a test tightly scoped to the bug to reproduce it and also demonstrate that it has been fixed?

I've been doing some experimentation on this, and it doesn't look like it would be _too_ much of a lift to add this. Let me know if you have questions, though! 👍

app/utils/findDocVersions.test.ts
```diff
@@ -0,0 +1,94 @@
...
+		expect(result).toEqual(['1.0'])
+	})
+
+	it('should use the default docsPathsAllVersions data if no custom data is provided', () => {
+		const spy = vi.spyOn(console, 'error').mockImplementation(() => {})
+		const result = findDocVersions('nonexistentProduct', '/docs/path1')
+		expect(result).toEqual([])
+		expect(spy).toHaveBeenCalledWith(
+			'Product, nonexistentProduct, not found in docs paths',
+		)
```
rmainwork, line 76: I'm not going to hold the review up over this, but we _could_ potentially do a regex here to make the error a little more fuzzy in case we ever change the error message verbiage or punctuation going forward...

I'll leave that to your judgement on whether or not to do that though.

```suggestion
		expect(spy.mock.calls[0][0]).toMatch(/nonexistentProduct.*not found/)
```

rmainwork (approved):

RubenSandwich (approved):
Double checked logs to make sure that returned content was the same and the `JSON.stringify` was not needed, and everything looked good. 👍🏻

PR #396, opened 2025-05-12
Compress upload into gzipped tarball

rmainwork:
This resolves an issue that was breaking preview deploys where too many files were being uploaded at once by compressing the entire upload into an archive prior to upload.

---
  - https://app.asana.com/0/0/1210228170623029

### DIFF
--- a/.github/workflows/build-pr-preview.yml
+++ b/.github/workflows/build-pr-preview.yml
@@ -7,7 +7,6 @@ on:
     branches:
       - main
       - develop
-      - rm/upload-tgz-archive-to-vercel # @todo - delete this
   # paths:
   # Hello Security 👋, we are checking to make sure forked repo PR changed paths are only in content/** inside the job security-check.
   # We are doing this so we can also reuse this workflow for internal PRs, as pull_request_target also triggers on internal PRs. (As does pull_request)


.github/workflows/build-pr-preview.yml
```diff
@@ -137,7 +137,7 @@ jobs:
         id: unified_docs_preview_url
         # save preview url from deployment and save copy of the preview url w/o https to fetch the deployment via Vercel REST API in next step
         run: |
-          preview_url=$(vercel deploy --prebuilt --target=preview --token=${{ secrets.VERCEL_TOKEN }})
+          preview_url=$(vercel deploy --prebuilt --target=preview --token=${{ secrets.VERCEL_TOKEN }}) --archive=tgz
```
LeahMarieBush, line 140: I think the archive=tgz tag should be within the parentheses
```suggestion
          preview_url=$(vercel deploy --prebuilt --target=preview --token=${{ secrets.VERCEL_TOKEN }} --archive=tgz)
```

rmainwork: Woah you're totally right, it should!

LeahMarieBush (commented):
Checked the preview and it looks like images are not working as expected (See [here](https://unified-docs-frontend-preview-rhmjzzlgd-hashicorp.vercel.app/terraform/cdktf)). However, when I inspected and clicked the image src link it works 🤔

rmainwork:
That's very odd especially since this change [hasn't actually gone into effect yet](https://github.com/hashicorp/web-unified-docs/actions/runs/14981320202/job/42085903921#step:11:2) (i.e: the version of the action that ran for this PR is the one on `main` since it's triggered by `pull_request_target`). I wonder if there's something else going on we need to look into...

rmainwork:
Did some digging into this today and it looks like it's actually due to the way that next loads images. I didn't realize that `next/image` basically proxies image requests. So if you right click the image and open it in another tab, it loads because you're logged into Vercel and have the relevant cookie. However, when `next/image` tries to load the image, it's making that request unauthenticated and so the image loading fails (hence the 401 in the browser network tab). I confirmed this suspicion by (temporarily) turning deployment protection off and reloading the page - at which point the image loaded as normal. I've also confirmed this affects other preview deployments too.

I don't think it's a big lift to fix this, but I've created [another ticket to address this](https://app.asana.com/1/90955849329269/project/1209314279859096/task/1210252710788466?focus=true) to keep things separate.

LeahMarieBush (commented):
I think we should also add this to the deploy-udr.yml file so that prod builds won't have this error.

Also, I am hesitant to approve this without the script actually running in the preview links. Is there a way you can run the new version of workflow so that I can test it?

rmainwork:
Looks like that worked. https://github.com/hashicorp/web-unified-docs/actions/runs/15028105571/job/42233831041#step:11:1

LeahMarieBush (approved):
Sweet LGTM! Thanks for adding a testing PR as well 🎉 Don't forget to remove the branch name on line 10

rmainwork:
Good point! Updated that in 32e500e. 

I'll wait for this CI to pass then merge.

PR #397, opened 2025-05-13
Updates for tfc-agent release 1.22.2

claire-tfc:
[Jira](https://hashicorp.atlassian.net/browse/TFDN-892)

### DIFF
--- a/content/terraform-docs-agents/v1.22.x/docs/cloud-docs/agents/agents.mdx
+++ b/content/terraform-docs-agents/v1.22.x/docs/cloud-docs/agents/agents.mdx
@@ -247,6 +247,7 @@ After initiating a graceful shutdown by either of these methods, the terminal us
 
     Environment variable: `TFC_AGENT_ACCEPT`
 
+@include 'policy.mdx'
 
 * `-request-forwarding`:
     Enable handling of forwarded HTTP requests. Enable this option only if you 


content/terraform-docs-agents/v1.22.x/docs/cloud-docs/agents/agents.mdx
```diff
@@ -247,7 +247,6 @@ After initiating a graceful shutdown by either of these methods, the terminal us
 
     Environment variable: `TFC_AGENT_ACCEPT`
 
-@include 'policy.mdx'
```
lafentres, line 250: idk why this automatically removed itself but i think we wanna keep it

lafentres (changes requested):
one small change then this should be good to go

content/terraform-docs-agents/v1.22.x/docs/cloud-docs/agents/agents.mdx
claire-tfc: That screenshot actually shows that the text is in here just below the include. Seems like it is removing a dup of content and we should keep this.

lafentres: it was recently added so i think if you wanna remove it you're gonna need to talk to the folks who added it first https://github.com/hashicorp/web-unified-docs-internal/pull/96

PR #346, opened 2025-04-15
Add `terraform-plugin-testing@1.13.x` documentation

bbasata:
What's changed from 1.12.x in this pull request?

```
$ git diff --stat 9cc15124d01e2e536254df46f2ac91ad51de52b7..HEAD content/terraform-plugin-testing
 content/terraform-plugin-testing/v1.13.x/data/plugin-testing-nav-data.json                                         |  30 +++++-
 content/terraform-plugin-testing/v1.13.x/docs/plugin/testing/acceptance-tests/continuous-integration.mdx           | 149 ++++++++++++++++++++++++++
 content/terraform-plugin-testing/v1.13.x/docs/plugin/testing/acceptance-tests/environment-variables.mdx            | 104 ++++++++++++++++++
 content/terraform-plugin-testing/v1.13.x/docs/plugin/testing/acceptance-tests/index.mdx                            | 315 ++++++++++++++++---------------------------------------
 content/terraform-plugin-testing/v1.13.x/docs/plugin/testing/acceptance-tests/testcase.mdx                         |   6 +-
 content/terraform-plugin-testing/v1.13.x/docs/plugin/testing/acceptance-tests/teststep/import-mode.mdx             | 382 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 content/terraform-plugin-testing/v1.13.x/docs/plugin/testing/acceptance-tests/{teststep.mdx => teststep/index.mdx} |  91 ++++++++++++----
 content/terraform-plugin-testing/v1.13.x/docs/plugin/testing/index.mdx                                             |  66 +++++-------
 content/terraform-plugin-testing/v1.13.x/docs/plugin/testing/testing-patterns.mdx                                  |  56 ++++++++++
```

1. terraform-plugin-testing 1.13.x adds new features to the "import mode" of acceptance tests. Because import mode is underdocumented, this pull request introduces `acceptance-tests/import-mode.mdx` and links it from `acceptance-tests/index.mdx`.

2. Refresh `acceptance-tests/index.mdx` so that "import mode" falls into an intuitive structure. It's a lengthy introduction, so this PR moves `environment-variables` and `continuous-integration` from subsections into separate pages.

   The info hierarchy is: testing -> acceptance testing -> test steps -> import mode.

3. Refresh & expand `acceptance-tests/teststep.mdx` so that "import mode" falls into an intuitive structure.

4. Refresh `index.mdx` to simplify.

5. Refresh `testing-patterns.mdx` to place import mode in context.

PR #401, opened 2025-05-16
Ignore gnu.org domain in link checker

RubenSandwich:
The link checker keeps timing out on gnu.org domains, likely due to our link checker being blocked as a bot, this PR just ignores checking the gnu.org domain links to help solve the problem of false positives.

https://github.com/hashicorp/web-unified-docs/pull/346#issuecomment-2851751679


RubenSandwich:
> This looks good - but I wonder if setting a custom user agent might also resolve this? I believe lychee [does support that](https://lychee.cli.rs/troubleshooting/network-errors/#try-a-different-user-agent)

I would feel strange about that, because they are obviously blocking bots. So I would rather respect their decision, plus they might then just block us in some other way and we would be stuck in a cat and mouse game. 🤷🏻

PR #436, opened 2025-05-28
Don't check compiled actions for missing copyright header

rmainwork:
The content of this action is(somewhat) auto generated by `ncc` meaning we don't have a great deal of control over its content. Therefore, we can probably safely ignore it when checking for missing copyright headers as it is isn't a source file


RubenSandwich (approved):
👍🏻

PR #440, opened 2025-05-28
Sync ptfe-releases and terraform-docs-common docs

rmainwork:
## 🗒️ What

Syncs terraform-docs-common and 

## 🤷 Why

Migration of terraform-docs-common and ptfe-releases to UDR.

## 💭 Anything else?

Of note: Due to the fact that #419 renames the `ptfe-releases` folder and product config to `terraform-enterprise`, it is no longer possible to use the migration script to migrate `terraform-enterprise`. Instead, I had to (temporarily) change the key in `productConfig.mjs` back to `ptfe-releases` so the script could locate the correct repo on GitHub, run the script (causing it to migrate the entire repo to my local machine) and then delete `content/terraform-enterprise` and rename `content/ptfe-releases` to `content/terraform-enterprise`. 

Just thought that was worth mentioning.


rmainwork:
After discussion with @RubenSandwich, we decided to go ahead and merge without fixing the link checker errors (since that's content and the link checker is new)

PR #456, opened 2025-06-01
change argument destroy to prevent_destroy

alifateh:
lifecycle { destroy = false } => wrong
lifecycle { prevent_destroy = true } => right


alifateh:
lifecycle { destroy = false } => wrong
lifecycle { prevent_destroy = true } => right

jbardin:
This is incorrect, the attribute name is `destroy`, which not work like the resource lifecycle `prevent_destroy`.

alifateh:
> This is incorrect, the attribute name is `destroy`, which does not work like the resource lifecycle `prevent_destroy`.

the Docs says prevent_destroy is correct attribute for change the default behaviour ...

`The arguments available within a lifecycle block are create_before_destroy, prevent_destroy, ignore_changes, and replace_triggered_by.`

please find it in link below
[https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#syntax-and-arguments](url)

alifateh:
> This is incorrect, the attribute name is `destroy`, which does not work like the resource lifecycle `prevent_destroy`.

Thanks, I confused between blocks [ resource - removed ] ;)

PR #473, opened 2025-06-09
Update v202505-1.mdx to include Podman warning

benemon:
As per discussion with engineering, add a warning to the current release page to highlight the potential discrepancy between our calculated container digest, and podman's reported digest.

### DIFF
--- a/content/terraform-enterprise/v202505-1/docs/enterprise/releases/2025/v202505-1.mdx
+++ b/content/terraform-enterprise/v202505-1/docs/enterprise/releases/2025/v202505-1.mdx
@@ -14,7 +14,7 @@ Flexible Deployment Options `terraform-enterprise` container digest: amd64/linux
 
 ```shell-session
 > podman pull images.releases.hashicorp.com/hashicorp/terraform-enterprise@sha256:4d31e2698e1ca2deb945a491d61f6bc914c0eb6e385e737a028b897a3eda4a63
-``
+```
 
 ## Highlights
 


benemon:
I thought it'd be good to include the actual pull command in the notification box, but it's messing up the formatting.

image

How would you prefer to handle this?

content/terraform-enterprise/v202505-1/docs/enterprise/releases/2025/v202505-1.mdx
```diff
@@ -10,6 +10,7 @@ Last required release: [v202406-1 (776)](/terraform/enterprise/releases/2024/v20
 
 Flexible Deployment Options `terraform-enterprise` container digest: amd64/linux `sha256:4d31e2698e1ca2deb945a491d61f6bc914c0eb6e385e737a028b897a3eda4a63`
 
+~> Users pulling images and validating digests with podman should note that there is an [outstanding issue](https://github.com/containers/podman/issues/25361) that may cause podman to misreport the container digest to the user. The recommended workaround until this is resolved is to pull by container digest directly. `podman pull images.releases.hashicorp.com/hashicorp/terraform-enterprise@sha256:4d31e2698e1ca2deb945a491d61f6bc914c0eb6e385e737a028b897a3eda4a63`
```
rkoron007, line 13: @benemon I'm going to play around with a few formatting options before I make a suggestion here! It'll make a minute since I can't build locally right now ([here is my preview PR](https://github.com/hashicorp/web-unified-docs/pull/474)), but wanted to let you know I'll make a formatting suggestion here before the day is out!

rkoron007: I looked into it more, and unfortunately we don't currently have a fix for the style breaking.

I agree having the command is super useful, and probably what folks want to see. Do you think it'd be enough to leave it at the top of the page without a callout? Maybe even adding an emoji since this is important and we do wanna catch the eye?

```suggestion
⚠️ If you are using Podman to deploy Terraform Enterprise, there is an [outstanding issue](https://github.com/containers/podman/issues/25361) that can cause Podman to misreport your container digest. In the meantime, we recommend pulling the container digest with `podman` directly:

```shell-session
> podman pull images.releases.hashicorp.com/hashicorp/terraform-enterprise@sha256:4d31e2698e1ca2deb945a491d61f6bc914c0eb6e385e737a028b897a3eda4a63
``
```

rkoron007: (I omitted the last backtick on code block in this suggestion so it wouldn't break the suggestion format -- if you like the suggestion you'd just need to add one more backtick at the bottom!

PR #483, opened 2025-06-11
fix: wait for UDR to be running before starting dev portal in make

LeahMarieBush:
### Description

This PR fixes terraform-docs-common redirects because there was a parsing error. It also updates the docker-compose file because the dev-portal image needs to wait for UDR to finish so it can access the redirects endpoint during it's build.

### Testing

Instructions are in [this PR](https://github.com/hashicorp/dev-portal/pull/2770)


RubenSandwich (approved):
🚢

PR #491, opened 2025-06-13
Link  docs accross versions using frontmatter ID

rmainwork:
This PR uses an `id` property in frontmatter to allow the linking of documents across versions. This is because the file URL is no longer a stable identifier (since by definition it changes if the file is moved or renamed) and allows the content authors control over _which_ documents they consider to be "the same" rather than trying to guess programmatically.


kaitlynnefuery:
@rmainwork Does this need to also be replicated in `dev-portal`? Looking at https://github.com/hashicorp/dev-portal/pull/2767 and presume it's outdated.

PR #492, opened 2025-06-13
Added comments about github and 3rd-party providers.

henryk-rychlik:
This wasn't explicitly spelled out previously. Only implied by the following sentence:
> Additionally, the agent must also be able to communicate with any services required by the Terraform code it is executing.


content/terraform-docs-agents/v1.22.x/docs/cloud-docs/agents/requirements.mdx
```diff
@@ -69,3 +69,5 @@ Additionally, the agent must also be able to communicate with any services requi
 | releases.hashicorp.com | tcp/443, HTTPS | Outbound       | Updating agent components and downloading Terraform binaries                                                                        |
 | archivist.terraform.io | tcp/443, HTTPS | Outbound       | Blob Storage   |
 | agents.terraform.io    | tcp/7146, TCP  | Outbound       | Agent RPC interface (currently used for request forwarding only) |
+
+Additionally, outbound traffic to *.github.com should be allowed if any 3rd-party providers are used. These can be pulled in implicitly by resource definitions, or explicitly by provider configurations added to the HCL code.
```
rkoron007, line 73: @henryk-rychlik Could you add this to the table above? I feel like it'd make sense with everything above! Something like this:

```
| github.com  | ? | Outbound       | Required if you are using a 3rd party provider while defining resources or in configuring providers. |
```

I'm not sure what the prot/protocal would be though?

PR #501, opened 2025-06-18
API Handler Function Wrapper

rmainwork:
Our tests contain quite a few references(and various competing implementations) of the following:

```ts
const mockRequest = (url: string) => {
  return new Request(url)
}
const request = mockRequest(`http://localhost:8080/api/all-docs-paths`)
const response = await GET(request)
```

I decided this could probably be streamlined a lot more by extracting all of those `mockRequest` calls into a single `callHandler` wrapper function which you can pass the API handler you're testing to. From there, it infers the param types (if the handler uses params) giving us automatic type hinting when testing API route handlers.

20250618_16h18m26s_grim
image

I do also just want to call out that last commit I added in at the end to consolidate (where appropriate) all the API handler parameters and their respective types. It's not _strictly_ in scope for this PR, but I was working in that general scope anyway and it wasn't too big of a lift, so I added it in.

---
  - https://app.asana.com/0/0/1210584876550952

### DIFF
--- a/content/terraform-docs-common/docs/cloud-docs/cost-estimation/index.mdx
+++ b/content/terraform-docs-common/docs/cloud-docs/cost-estimation/index.mdx
@@ -6,8 +6,6 @@ description: >-
 
 # Cost estimation overview
 
-> **Hands-on:** Try the [Control Costs with Policies](/terraform/tutorials/cloud-get-started/cost-estimation) tutorial to practice enabling cost estimation and define a policy to check the total monthly delta.
-
 HCP Terraform provides cost estimates for many resources found in your Terraform configuration. For each resource an hourly and monthly cost is shown, along with the monthly delta. The total cost and delta of all estimable resources is also shown.
 
 ## Enabling Cost Estimation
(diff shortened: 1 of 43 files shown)


app/utils/callHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+ * ```ts
+ *   import { GET } from './route.ts'
+ *   // ...
+ *   const response = await requestWrapper(GET, { product: 'terraform' })
+ *
+ *   expect(response.status).toBe(200)
+ *   // Other assertions...
+ * ```
+ */
+export const callHandler = <
```
RubenSandwich, line 22: @rmainwork let's give this a more description name, so folks know that this is used on;y in testing.

Something like `testRequestHandler` or `createTestRequestHandler`.

rmainwork: Agreed. I can't say I was crazy about the name, I just picked it to name it something. I ended up going with `mockRequest` in 1a37a8033992f3547bfc6edac1e2171b5119dae8. My reasoning was that it's the same name as the previous implementation, so it should hopefully nudge people toward using that instead of creating their own `mockRequest` functions...

I didn't end up going with `testRequestHandler` or `createTestRequestHandler` as they don't exactly test the handler themselves or create the handler which those names seem to imply.

I honestly don't _love_ the name `mockRequest` as it implies (at least to me) that the function somehow mocks part of the request stack out(I suppose it _sort_ of does but ehhhh) rather than just calls the handler provided in a predetermined way.

RubenSandwich (approved):
👍🏻

PR #504, opened 2025-06-19
Improve clarity on gitlab ee / ce webhook configurations

benemon:
As per discussion on Slack, moving the information about "Allow requests to the local network from hooks and service" in Gitlab self hosted options into troubleshooting to better alert the user as to why they would need to do this.

In discussions, it appears that a couple of folks have missed the significance of that section of the docs due to it not being explicitly related to the 422 error seen in the TFE UI.

This PR will remove that ambiguity by specifically calling out the error message and the resolution.


nikolasrieble (approved):
The wording is identical, what is added is a header in the troubleshooting section that provides details on the error message. This is great because it could help customers search for the issue more easily. 

Thank you! 

~~Please evaluate the targets for backporting.~~
EDIT: Upon reflection, I believe we do not need to backport this change. The change is merely stylistic / adding further context - which is great. But the information is already available in previous releases. 

So I suggest: Lets pragmatically not backport :)

PR #506, opened 2025-06-19
Remove callout to Cost Estimation tutorial in favor of the usage doc

BrianMMcClain:
Since the usage doc and tutorial cover the same content, we've decided to consolidate into the single usage doc. As such, we no longer need this callout to the tutorial.

### DIFF
--- a/content/terraform/v1.10.x/docs/intro/use-cases.mdx
+++ b/content/terraform/v1.10.x/docs/intro/use-cases.mdx
@@ -48,7 +48,7 @@ Terraform can help you enforce policies on the types of resources teams can prov
 
 ### Resources
 
-- The [cost estimation documentation](/terraform/cloud-docs/cost-estimation) provides information to estimate the cost of infrastructure changes and define policies to limit it.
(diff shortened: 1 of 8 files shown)


content/terraform/v1.10.x/docs/intro/use-cases.mdx
```diff
@@ -48,7 +48,7 @@ Terraform can help you enforce policies on the types of resources teams can prov
 
 ### Resources
 
-- Try the [Control Costs with Policies](/terraform/tutorials/cloud-get-started/cost-estimation) tutorial to estimate the cost of infrastructure changes and define policy to limit it.
+- The [cost estimation documentation](/terraform/cloud-docs/cost-estimation) provides information to estimate the cost of infrastructure changes and define policies to limit it.
```
trujillo-adam, line 51:
```suggestion
- The [cost estimation documentation](/terraform/cloud-docs/cost-estimation) describes how to use the cost estimation feature and how to define policies that limit costs associated with infrastructure changes.
```
It might be silly, but the original wording sounds as if the docs actually has cost estimation and policy data.

trujillo-adam (approved):
I added one suggestion to clarify the language a bit. It looks like it's the same statement copied to the other versions of the page, but I only added it once. Feel free to tweak as necessary.

PR #526, opened 2025-07-01
Add End of Sale note for HCP Vault Secrets

rkoron007:
Added End of Sale note to HCP TF and TFE v202411 and up (this feature was introduced in 202411). I also visually confirmed the deprecation note across each TFE version!

Jira ticket: https://hashicorp.atlassian.net/browse/IPE-1195

### DIFF
--- a/content/terraform-docs-common/docs/docs/partnerships.mdx
+++ b/content/terraform-docs-common/docs/docs/partnerships.mdx
@@ -159,8 +159,11 @@ The external system can then use this run information and respond back to HCP Te
 
 Partners who successfully complete the HCP Terraform integration checklist obtain an HCP Terraform badge. This badge signifies HashiCorp validated the integration under ideal conditions and the partner is a member of the HashiCorp Technology Partner Program.
 
+<ImageConfig height={200} width={200} hideBorder>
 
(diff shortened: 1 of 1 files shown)


content/terraform-docs-common/docs/partials/deprecation/hcp-vault-secrets.mdx
```diff
@@ -0,0 +1,8 @@
+<Highlight title="End of sale announcement">
+
+HCP Vault Secrets will no longer be available for purchase by new customers after June 30th, 2025.
```
robin-norwood, line 3: Maybe over thinking it, but does "by new customers" possibly imply that existing HashiCorp customers could still purchase it after? Would `HCP Vault Secrets will no longer be available for purchase after June 30th, 2025.` be more direct?

robin-norwood (approved):
A question about the wording of the annoucement, but pre-approving!

content/terraform-docs-common/docs/partials/deprecation/hcp-vault-secrets.mdx
rkoron007: Ooh yes! I like that! I'm going to change it everywhere!

PR #535, opened 2025-07-02
normalize blocks nav

trujillo-adam:
This PR makes the reference page titles and configuration blocks nav consistent. IMO, we should use lower case and remove the "block" at the end so that it more closely matches `{block-type}`, but I'll yield if others disagree.


robin-norwood (approved):
lgtm!

PR #542, opened 2025-07-03
clarify permission for tag mgmt requirements

drewmullen:
It is possible to manage tags on workspaces with organizational access `manage all workspaces`. Calling that out in the docs.

Please let @hashicorp/technical-program-mgmt know if you have any questions.


trujillo-adam (commented):
Thanks for the update. I made some suggestions per the style guide.

PR #552, opened 2025-07-09
terraform: fix reported broken links

radeksimko:
As reported in https://github.com/hashicorp/web-unified-docs/pull/551#issuecomment-3052907187


SarahFrench (approved):
LGTM and looks like the last 404 reported by automation was fixed

PR #605, opened 2025-07-22
Fix TFE_REDIS_USE_MTLS header level in configuration reference

tism:
TFE_REDIS_USE_MTLS had a title level one higher than the other options

image


nikolasrieble (approved):
Thank you for the quick fix!

PR #632, opened 2025-07-24
fix: set failIsEmpty to false for link checker

LeahMarieBush:
### Description

This PR sets the `failIfEmpty` parameters to false for the lychee link checker github action. This makes it so if a file without any links is changed it won't fail that step of the github action. You can see this happening in [this github action run](https://github.com/hashicorp/web-unified-docs/actions/runs/16484005624/job/46608005580?pr=627)

### Testing

There's not a great way to test this since the updates for this action won't go into effect until this is merged into main


Younique98 (approved):
Looks great! Good idea!

PR #655, opened 2025-07-30
Add documentation for new vault secret list parameter

Shaeli:
Documentation for this change: https://github.com/hashicorp/vault-plugin-secrets-kv/pull/196

cc @robmonte for helping me with this new small feature ✨ 

 cc https://github.com/github/glb/issues/7601

### DIFF
--- a/content/vault/v1.21.x (rc)/content/api-docs/secret/kv/kv-v2.mdx
+++ b/content/vault/v1.21.x (rc)/content/api-docs/secret/kv/kv-v2.mdx
@@ -531,7 +531,7 @@ the secret to list, such as `secret`. This is specified as part of the URL.
 $ curl \
     --header "X-Vault-Token: ..." \
     --request LIST \
-    https://127.0.0.1:8200/v1/secret/metadata/my-secret?exclude_deleted=false
+    https://${VAULT_ADDR}/v1/secret/metadata/my-secret?exclude_deleted=false
 ```
 
 ### Sample response


robmonte (approved):
Thanks for taking care of this!

Shaeli:
@robmonte I cannot merge de PR - I'll let you click the button whenever you're ready on your side to release this feature.
thanks again for the help getting this 🚢 ✨

content/vault/v1.21.x/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -523,13 +523,15 @@ the secret to list, such as `secret`. This is specified as part of the URL.
 - `path` `(string: <required>)` – Specifies the path of the secrets to list.
   This is specified as part of the URL.
 
+- `exclude_deleted` `(bool: false)` – If set to true, only non-deleted entries will be returned.
```
schavis, line 526:
```suggestion
- `exclude_deleted` `(bool: false)` – Only return non-deleted entries.
```

content/vault/v1.21.x/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -523,13 +523,15 @@ the secret to list, such as `secret`. This is specified as part of the URL.
...
+
 ### Sample request
 
 ```shell-session
 $ curl \
     --header "X-Vault-Token: ..." \
     --request LIST \
-    https://127.0.0.1:8200/v1/secret/metadata/my-secret
+    https://127.0.0.1:8200/v1/secret/metadata/my-secret?exclude_deleted=false
```
schavis, line 534:
```suggestion
    https://${VAULT_ADDR}/v1/secret/metadata/my-secret?exclude_deleted=false
```
We're slowly converting example code to use the `VAULT_ADDR` environment variable instead of hardcoding a localhost address

schavis (changes requested):

content/vault/v1.21.x (rc)/content/api-docs/secret/kv/kv-v2.mdx
```diff
@@ -523,13 +523,15 @@ the secret to list, such as `secret`. This is specified as part of the URL.
...
+- `exclude_deleted` `(bool: false)` – Only return non-deleted entries.
+
 ### Sample request
 
 ```shell-session
 $ curl \
     --header "X-Vault-Token: ..." \
     --request LIST \
-    https://127.0.0.1:8200/v1/secret/metadata/my-secret
+    https://${VAULT_ADDR}/v1/secret/metadata/my-secret?exclude_deleted=false
```
robmonte, line 534: I was thinking, I don't know if we provide examples in-line to the URL like this anywhere else within the docs. Do you know if you've seen this before? 

It might match our normal convention better to provide json of the field and pass it in as a data payload.

```
### Sample payload

```json
{
  "exclude_deleted": true
}
```

and then putting 
`--data @payload.json \`
within the example request. Also perhaps showing setting to true makes for a better example.

robmonte: Oh that's a good point. Besides, this is a preexisting example of a path parameter instead of payload. I think it's fair to leave it as-is then.

schavis (approved):

PR #659, opened 2025-07-30
docs: enable Explorer API docs for TFE

notchairmk:
Enables API docs for Explorer for TFE

https://hashicorp.atlassian.net/browse/TF-26731


notchairmk:
🙂 either way works for me

PR #668, opened 2025-07-31
Update title() function documentaton to show non-working example.

bschonec:
I wrongly assumed that the title() function would lower-case characters.  Perhaps this small addition will make the function clearer.

### DIFF
--- a/content/terraform/v1.12.x/docs/language/functions/title.mdx
+++ b/content/terraform/v1.12.x/docs/language/functions/title.mdx
@@ -7,7 +7,7 @@ description: |-
 
 # `title` Function
 
-`title` converts the first letter of each word in the given string to uppercase.  It does not downcase any letters in the string.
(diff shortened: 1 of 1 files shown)


crw:
I was able to reproduce this behavior. 👍

content/terraform/v1.12.x/docs/language/functions/title.mdx
```diff
@@ -7,14 +7,18 @@ description: |-
 
 # `title` Function
 
-`title` converts the first letter of each word in the given string to uppercase.
+`title` converts the first letter of each word in the given string to uppercase.  It does not downcase any letters in the string.
```
rkoron007, line 10: Hello @bschonec! Great catch and thank you for the contribution! How would you feel about rewording this a little:
```suggestion
`title` converts the first letter of each word in the given string to uppercase. The `title` function does not change any other letters in the string.
```

Thanks for adding an example too!

rkoron007 (approved):
Looks great! Added a small suggestion for clarity!

PR #669, opened 2025-07-31
Update HYOK Configurations Nav Title

dominic-retli-hashi:
Changed from `Key Configurations` to `Configurations` as it should have been originally for nav to [this page](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/hold-your-own-key/configurations).

JIRA: https://hashicorp.atlassian.net/browse/TF-26977


rkoron007 (approved):
Sorry, thank you for catching this!!

PR #673, opened 2025-08-01
Update CODEOWNERS so that changes to content require education approval.

Younique98:
Contributors were not being blocked properly. Content changes require an education team members approval.

The existing rule does not cover files under that deeper nested path.


Younique98:
Testing

PR #675, opened 2025-08-04
it needs an empty line

munnep:
IBM verify page needs an empty line


munnep:
A table isn't shown properly. I verified it locally and it works but on the page itself it looks weird. I think an empty line is needed to make sure it actually renders as a table

nikolasrieble (approved):

PR #676, opened 2025-08-04
Vault: Add Secure cluster with TLS to Concepts for all versions

brianshumate:
Convert secure with TLS content from WAF to documentation and add to Concepts for all versions.


content/vault/v1.20.x/content/docs/concepts/secure-cluster-with-tls.mdx
```diff
@@ -0,0 +1,49 @@
+---
+layout: docs
+page_title: Secure cluster with TLS
+description: >-
+  Protect cluster communications and security posture with end-to-end TLS.
+---
+
+When you secure your Vault cluster communications with TLS, you enhance the cluster's overall security posture. TLS also ensures that data transmitted between Vault nodes and clients remains confidential and tamper-proof.
```
yhyakuna, line 8: For docs, we need to set the page title explicitly.
image


```suggestion
# Secure cluster with TLS

When you secure your Vault cluster communications with TLS, you enhance the cluster's overall security posture. TLS also ensures that data transmitted between Vault nodes and clients remains confidential and tamper-proof.
```

schavis (approved):

PR #677, opened 2025-08-04
feat: sort versions by semver first and then non-semver

LeahMarieBush:
### Description

This PR sorts versions differently. When there is semver and non-semver versions within a product (terraform-enterprise for example), the semver versions are getting sorted to the bottom of the list. Since these versions are more recent, they should be at the top of the list. This PR fixes that by sorting the semver and non-semver separately and then combining them into one version list. 
This PR has a [partner PR](https://github.com/hashicorp/dev-portal/pull/2807) in the dev-portal repo that allows for versions without a 'v' at the beginning.

### Testing

This PR includes changes to dev-portal to allow for versions without a 'v'. You will need to locally run UDR and dev-portal to see these changes reflected. 

First, do the steps in UDR:
UDR: 
1. Checkout `leah/feat/sort-versions`
2. Go to terraform-enterprise and create a few new versions. You can do this by copying the latest version folder and pasting it and then changing the name. For testing the sorting, use a variety of version names (ex. 1.0.0, 1.0.1, 1.1.0, 9.9.9)
3. Run `npm run prebuild`
4. After that finishes, run `npm run dev`

Then do the steps in dev-portal:
Dev-portal: 
1. Checkout `leah/feat/versions-without-v`
2. In the .env file, change the `UNIFIED_DOCS_API` to `http://localhost:8080`
3. Run `npm start`

### DIFF
--- a/scripts/gather-version-metadata.test.ts
+++ b/scripts/gather-version-metadata.test.ts
@@ -169,3 +169,68 @@ it('Sort order - 1.10 is higher than 1.9', async () => {
 	const result = await gatherVersionMetadata('/content')
 	expect(result).toStrictEqual(expected)
 })
+
+it('handles semver versions', async () => {
+	const expected = {
+		'terraform-enterprise': [
+			{ version: '1.2.0', releaseStage: 'stable', isLatest: true },
+			{ version: '1.1.0', releaseStage: 'stable', isLatest: false },
+			{ version: '1.0.0', releaseStage: 'stable', isLatest: false },
+		],
+	}
+	vol.fromJSON(
+		{
+			'./terraform-enterprise/1.2.0/': null,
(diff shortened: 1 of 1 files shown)


scripts/gather-version-metadata.test.ts
```diff
@@ -169,3 +169,68 @@ it('Sort order - 1.10 is higher than 1.9', async () => {
...
+			'./terraform-enterprise/v202506-01/': null,
+		},
+		// default cwd
+		'/content',
+	)
+	const result = await gatherVersionMetadata('/content')
+	expect(result).toStrictEqual(expected)
+})
+
+it('handles mixed semver and non-semver versions', async () => {
```
RubenSandwich, line 215: ❤️

PR #680, opened 2025-08-04
replace  with

trujillo-adam:
This PR replaces the `provider` meta-argument in the `module` block with the correct `providers` meta-argument.

### DIFF
--- a/content/terraform/v1.12.x/docs/language/block/module.mdx
+++ b/content/terraform/v1.12.x/docs/language/block/module.mdx
@@ -502,7 +502,7 @@ module "<LABEL>" {
 }
 ```
 
-By default, the child module inherits the default provider configuration from its parent module, but you can create multiple provider configurations and use an alternate configuration for each module. Because a module is collection of different resources, some modules also let you specify a alternate configuration for different resources in the module. Refer to the module documentation for details.
+By default, a child module inherits the default provider configuration from its parent module, but you can create multiple provider configurations and use an alternate configuration for each module. Because a module is collection of different resources, some modules also let you specify an alternate configuration for different resources in the module. Refer to the module documentation for details.
 
 The `providers` argument is a map of key-value pairs that reference provider configurations. The key references the provider configuration name used in the child module. Refer to the module documentation for details on how to define keys. The value references the provider configuration name from the parent module. Use the `<PROVIDER>.<ALIAS>` syntax to reference alternate provider configurations in key values.  
 


content/terraform/v1.12.x/docs/language/block/module.mdx
```diff
@@ -490,27 +490,29 @@ The `for_each` argument is a meta-argument, which is built into Terraform and co
...
 module "<LABEL>" {
-  provider = <provider>.<alias>
+  providers = {
+   <provider> = <provider>.<alias>
+  }
 }
 ```
 
-By default, Terraform applies the default provider based on the module resource type, but you can create multiple provider configurations and use a non-default configuration for specific modules. 
+By default, the child module inherits the default provider configuration from its parent module, but you can create multiple provider configurations and use an alternate configuration for each module. Because a module is collection of different resources, some modules also let you specify a alternate configuration for different resources in the module. Refer to the module documentation for details.
```
rkoron007, line 505: I think you meant to add a link too at the end of this sentence?
```suggestion
By default, a child module inherits the default provider configuration from its parent module, but you can create multiple provider configurations and use an alternate configuration for each module. Because a module is collection of different resources, some modules also let you specify an alternate configuration for different resources in the module. Refer to the module documentation for details.
```

rkoron007: Also a small dumb nit is that modules inherit default provider configurations expect for the `source` and `version` - you probably cover that somewhere else, but figured I'd ask it just in case.

content/terraform/v1.12.x/docs/language/block/module.mdx
```diff
@@ -490,27 +490,29 @@ The `for_each` argument is a meta-argument, which is built into Terraform and co
...
 
-Use the `<PROVIDER>.<ALIAS>` syntax to reference a provider configuration in the `provider` argument. Refer to [Multiple Provider Configurations](/terraform/language/providers/configuration#alias-multiple-provider-configurations) for instructions on how to reference a specific provider configuration.     
+The `providers` argument is a map of key-value pairs that reference provider configurations. The key references the provider configuration name used in the child module. Refer to the module documentation for details on how to define keys. The value references the provider configuration name from the parent module. Use the `<PROVIDER>.<ALIAS>` syntax to reference alternate provider configurations in key values.
```
rkoron007, line 507: Need a link here, too!

content/terraform/v1.12.x/docs/language/block/module.mdx
```diff
@@ -875,23 +877,26 @@ module "ec2_instance" {
...
-  provider = google.europe
-  # ...
+module "tunnel" {
+  source    = "./tunnel"
+  providers = {
+    aws.src = aws.usw1
+    aws.dst = aws.usw2
+  }
 }
 ```
```
rkoron007, line 902: I think it would be cool to expand this example a little further by showing what the config of the `tunnel` module looks like— helping people connect the dots of how they can use the `provider` argument in a resource of the `tunnel` module to reference the provider configurations passed in through the `providers` argument. 

But if you feel like that is too out of scope that makes sense too!

rkoron007 (approved):

PR #689, opened 2025-08-05
Load redirects from content dir

RubenSandwich:
## Ticket
[Redirects for local build are not pulling from local UDR API](https://app.asana.com/1/90955849329269/project/1210146125607822/task/1210952396134799?focus=true)

## Description

During local development we don't want to use the API to serve redirect files from `/public`, rather we want to read the redirect file in `/content`.

## Testing
**Local Dev**

1. Run locally `npm run dev`
    1. prebuild is not needed
1. Edit a redirect file, such as:
    1. `content/vault/v1.20.x/redirects.jsonc`
1. Go to the redirect API endpoint for that product to confirm your changes
    1. http://localhost:8080/api/content/vault/redirects
    
**Preview/Production**

1. Make sure the redirect API is still working as intended:
    1. [/api/content/vault/redirects](https://web-unified-docs-g7vh3mdx1-hashicorp.vercel.app/api/content/vault/redirects)
1. `process.env.NODE_ENV` in vercel is always "production" when deployed on vercel's servers, so "development" should only ever happen locally

### DIFF
--- a/app/api/content/[productSlug]/redirects/route.ts
+++ b/app/api/content/[productSlug]/redirects/route.ts
@@ -40,7 +40,7 @@ export async function GET(request: Request, { params }: { params: GetParams }) {
 	]
 
 	const readFileResult = await findFileWithMetadata(filePath, versionMetadata, {
-		loadFromContentDirNotPublic: process.env.VERCEL_ENV === 'development',
+		loadFromContentDir: process.env.NODE_ENV === 'development',
 	})
 	if (!readFileResult.ok) {
 		return new Response('Not found', { status: 404 })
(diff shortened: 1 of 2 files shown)


LeahMarieBush (commented):
I ran this locally and I'm still running into the issue where the local redirects are not being used. Here are the steps I took: 
1. Changed a redirect in one of the redirects files
2. Ran `make`
3. Went to the link I changed and it is still redirecting to the old link

LeahMarieBush (approved):
Nvm, my redirect was cached so changing it did not work. LGTM!

PR #691, opened 2025-08-05
VSO CSI driver documentation

digivava:
This PR contains documentation for a new feature being added to the VSO Helm chart, a HashiCorp-maintained "CSI driver". 

I also updated other relevant pages to link to it for easier discoverability.

### DIFF
--- a/content/vault/v1.20.x/data/docs-nav-data.json
+++ b/content/vault/v1.20.x/data/docs-nav-data.json
@@ -904,7 +904,7 @@
                 "path": "deploy/kubernetes/vso/helm"
               },
               {
-                "title": "Direct secrets",
+                "title": "Use VSO with CSI",
                 "badge": {
                   "text": "ENT",
                   "type": "filled",
@@ -951,7 +951,7 @@
                 "path": "deploy/kubernetes/csi/installation"
               },
               {
-                "title": "Configurations",
+                "title": "Configuration",
                 "path": "deploy/kubernetes/csi/configurations"
               },
               {


content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/helm.mdx
```diff
@@ -39,6 +40,8 @@ Use these links to navigate to a particular top-level stanza.
 
   - `replicas` ((#v-controller-replicas)) (`integer: 1`) - Set the number of replicas for the operator.
 
+  - `priorityClassName` ((#v-controller-priorityclassname)) (`string: ""`) - Set the priority class for the operator.
```
digivava, line 43: `priorityClassName`, `topologySpreadConstraints`, and `podDisruptionBudget` are not part of this work, but they got auto-generated along with the rest of this because it looks like docs generation got missed last time.

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
digivava, line 1: I think I'll need your review on this file @stevealmyHC to determine if modifying this existing page is sufficient for satisfying the comparison ticket for now, or if you want to go further and create a dedicated page for the comparisons. (This is currently just on the "Kubernetes" overview page.)

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
```diff
@@ -1,14 +1,17 @@
...
-  Learn how the Vault CSI provider allows pods to consume Vault secrets using CSI volumes.
+  Learn how the Vault provider for the Secrets Store CSI Driver allows pods to consume Vault secrets using CSI volumes.
 ---
 
-# Vault CSI provider
+# Vault CSI provider for the Secrets Store CSI Driver
 
 The Vault CSI Provider allows pods to consume Vault secrets using
-[CSI Secrets Store](https://github.com/kubernetes-sigs/secrets-store-csi-driver) volumes.
+[Secrets Store CSI](https://github.com/kubernetes-sigs/secrets-store-csi-driver) volumes.
```
schavis, line 11:
```suggestion
[Secrets Store CSI](https://github.com/kubernetes-sigs/secrets-store-csi-driver)
volumes. You can deploy the
[HashiCorp CSI driver](/vault/docs/platform/k8s/vso/vso-csi) as part of your
Vault Secrets Operator instance.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
```diff
@@ -1,14 +1,17 @@
...
+
+~> For HashiCorp's CSI driver which can be deployed as part of the Vault Secrets Operator, 
+see the documentation for the [Vault Secrets Operator CSI driver](/vault/docs/platform/k8s/vso/vso-csi).
```
schavis, line 14:
```suggestion
```
There's no need to make this a separate alert. We can fold it into the introduction

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,22 +57,32 @@ There are several ways to try Vault with Kubernetes in different environments.
 
 ### High level comparison of integrations
 
-There are currently 3 different integrations to help Kubernetes workloads seamlessly consume secrets from Vault, without the need to modify the application to interact directly with Vault. Each integration addresses slightly different use-cases. The following is a brief overview of the strengths of each integration.
+There are currently 4 different integrations to help Kubernetes workloads seamlessly consume secrets from Vault, without the need to modify the application to interact directly with Vault. Each integration addresses slightly different use-cases. The following is a brief overview of the strengths of each integration.
 
 #### Agent injector
+([docs](/vault/docs/platform/k8s/injector))
```
schavis, line 63:
```suggestion

([Agent injector docs](/vault/docs/platform/k8s/injector))
```
Style correction: use meaningful anchor text

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,22 +57,32 @@ There are several ways to try Vault with Kubernetes in different environments.
...
 
 - No durable secret storage outside Vault. All secrets written to disk are in ephemeral in-memory volumes.
 - No highly privileged service accounts required. All secrets are fetched with the pod's own service account without the need for any other service accounts to impersonate it.
-- More mature solution, with proven production record and advanced features like templating,
-  wider array of auth methods, etc.
+- Wider array of auth methods
 
 #### Vault Secrets Operator
+([docs](/vault/docs/platform/k8s/vso))
```
schavis, line 70:
```suggestion

([Vault Secrets Operator docs](/vault/docs/platform/k8s/vso))
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,22 +57,32 @@ There are several ways to try Vault with Kubernetes in different environments.
...
 
 - More native UX for app developers. Workloads can mount Kubernetes secrets without adding any Vault-specific configuration.
 - Reduced load on Vault. Secrets are synced per CRD instead of per consuming pod.
 - Better Vault secret availability. Kubernetes secrets act as a durable cluster-local cache of Vault secrets.
+- Ability to use templating through [SecretTransformations](/vault/docs/platform/k8s/vso/secret-transformation) to deliver secrets in the desired format
```
schavis, line 75:
```suggestion
- Supports templating with [SecretTransformations](/vault/docs/platform/k8s/vso/secret-transformation)
  that deliver secrets in the desired format.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,22 +57,32 @@ There are several ways to try Vault with Kubernetes in different environments.
...
 
-#### Vault CSI provider
+#### Vault Secrets Operator CSI driver
+([docs](/vault/docs/platform/k8s/vso/vso-csi))
```
schavis, line 78:
```suggestion

([CSI driver for Vault Secrets Operator docs](/vault/docs/platform/k8s/vso/vso-csi))
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,22 +57,32 @@ There are several ways to try Vault with Kubernetes in different environments.
...
+
+- All secrets are written only to ephemeral volumes, for users who wish to avoid the locked-down API server permissions and etcd encryption requirements of Kubernetes Secrets.
```
schavis, line 80:
```suggestion
- Writes all secrets to ephemeral volumes to avoid the Kubernetes Secrets
  requirements for locked-down API server permissions and `etcd` encryption.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,22 +57,32 @@ There are several ways to try Vault with Kubernetes in different environments.
...
+- Reduced load on Vault compared to Agent Injector, as the number of connections to Vault corresponds to the number of nodes, not the number of consuming pods.
```
schavis, line 81:
```suggestion
- Associates Vault connections with nodes instead of consuming pods, which
  reduces your Vault load when compared to Agent Injector.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,22 +57,32 @@ There are several ways to try Vault with Kubernetes in different environments.
...
+- Ability to use templating through [SecretTransformations](/vault/docs/platform/k8s/vso/secret-transformation) to deliver secrets in the desired format
```
schavis, line 82:
```suggestion
- Supports templating with [SecretTransformations](/vault/docs/platform/k8s/vso/secret-transformation)
  that deliver secrets in the desired format.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,22 +57,32 @@ There are several ways to try Vault with Kubernetes in different environments.
...
+
+#### Vault CSI provider for Secrets Store CSI driver
+([docs](/vault/docs/platform/k8s/csi))
```
schavis, line 85:
```suggestion

([Vault CSI provider for Secrets Store CSI driver docs](/vault/docs/platform/k8s/csi))
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/index.mdx
```diff
@@ -19,6 +19,10 @@ The operator writes the *source* secret data directly to the *destination* Kuber
 changes made to the *source* are replicated to the *destination* over its lifetime. In this way, an application only needs
 to have access to the *destination* secret in order to make use of the secret data contained within.
 
+For Vault Enterprise users, the Vault Secrets Operator can be alternatively deployed alongside a CSI driver to mount secrets directly to pods, 
+without making use of Kubernetes Secrets. The rest of this page covers the traditional operation of Vault Secrets Operator. 
+You can learn more about the CSI driver option on the [Vault Secrets Operator CSI driver](/vault/docs/platform/k8s/vso/vso-csi) documentation page.
```
schavis, line 24:
```suggestion

<Tip>

Vault Enterprise users can deploy the Vault Secrets Operator alongside a CSI
driver to mount secrets directly to pods without using Kubernetes Secrets.

Refer to the
[Vault Secrets Operator CSI driver](/vault/docs/platform/k8s/vso/vso-csi)
documentation to learn more about the CSI driver option.

</Tip>

```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/installation.mdx
```diff
@@ -41,6 +41,17 @@ Then install the Operator:
 $ helm install --version 0.10.0 --create-namespace --namespace vault-secrets-operator vault-secrets-operator hashicorp/vault-secrets-operator
 ```
 
+## CSI driver option
+
+The Vault Secrets Operator Helm chart can also be used to deploy a DaemonSet of the [Vault Secrets Operator CSI driver](/vault/docs/platform/k8s/vso/vso-csi). 
+To do so, add the flag `"csi.enabled=true"`.
```
schavis, line 47:
```suggestion
You can use Vault Secrets Operator Helm chart to deploy a DaemonSet of the
[Vault Secrets Operator CSI driver](/vault/docs/platform/k8s/vso/vso-csi) with
the `"csi.enabled"` flag:
```

schavis (changes requested):

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
digivava: Discussed offline, sounds like we will add a more detailed comparison as well.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/helm.mdx
digivava: The Helm docs are generated by the `make gen-helm-docs` command in the [vault-secrets-operator](https://github.com/hashicorp/vault-secrets-operator) repo. (Well, this particular iteration is from a private fork of that repo, but that's typically where the generation is done.)

Is the Education team typically added as a codeowner for field descriptions in the main Vault repo, like wherever the Vault CLI help text lives, for example? I would say this is similar to that -- the text lives with the code itself, but we have a helper script that spits it out in this format.

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
schavis: We want to avoid overloading pages with asides whenever possible, but a tip might make sense in this case. Let me think about it and add a suggestion tomorrow.

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/configurations.mdx
```diff
@@ -1,13 +1,13 @@
...
 layout: docs
-page_title: Vault CSI provider configurations
+page_title: Vault Secrets Store CSI Provider configurations
 description: >-
-   Learn about the command line arguments and the secret provider class parameters supported by the Vault CSI provider.
+   Learn about the command line arguments and the secret provider class parameters supported by the Vault Secrets Store CSI Provider.
 ---
 
-# Vault CSI provider configurations
+# Vault Secrets Store CSI Provider configurations
```
schavis, line 8:
```suggestion
# Vault Secrets Store CSI provider configurations
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/configurations.mdx
```diff
@@ -1,13 +1,13 @@
...
 
-The following command line arguments are supported by the Vault CSI provider.
+The following command line arguments are supported by the Vault Secrets Store CSI Provider.
```
schavis, line 10:
```suggestion
The following command line arguments are supported by the Vault Secrets Store CSI provider.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/configurations.mdx
```diff
@@ -38,8 +38,8 @@ If installing via the helm chart, they can be set using e.g.
   for connecting to Vault. Can also be specified via the `VAULT_ADDR` environment
   variable. **Note:** It is highly recommended to only set the Vault address when
   installing the helm chart. The helm chart will install Vault Agent as a sidecar
-  to the Vault CSI Provider for caching and renewals, but setting `-vault-addr`
-  here will cause the Vault CSI Provider to bypass the Agent's cache.
+  to the Vault Secrets Store CSI Provider for caching and renewals, but setting `-vault-addr`
+  here will cause the Vault Secrets Store CSI Provider to bypass the Agent's cache.
```
schavis, line 42:
```suggestion
  to the Vault Secrets Store CSI provider for caching and renewals. Setting `-vault-addr`
  here causes the Vault Secrets Store CSI provider to bypass the Vault Agent cache.
```
I'm assuming Vault Agent is the "Agent" mentioned in this section?

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/configurations.mdx
```diff
@@ -88,9 +88,8 @@ structure is illustrated in the [examples](/vault/docs/platform/k8s/csi/examples
 
 - `vaultAddress` `(string: "")` - The address of the Vault server. **Note:** It is
   highly recommended to only set the Vault address when installing the helm chart.
-  The helm chart will install Vault Agent as a sidecar to the Vault CSI Provider
-  for caching and renewals, but setting `vaultAddress` here will cause the Vault
-  CSI Provider to bypass the Agent's cache.
+  The helm chart will install Vault Agent as a sidecar to the Vault Secrets Store CSI Provider
+  for caching and renewals, but setting `vaultAddress` here will cause the Vault Secrets Store CSI Provider to bypass the Agent's cache.
```
schavis, line 92:
```suggestion
  The helm chart installs Vault Agent as a sidecar to the Vault Secrets Store CSI provider
  for caching and renewals. Setting `vaultAddress` here causes the Vault Secrets Store CSI provider to bypass the Vault Agent cache.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
```diff
@@ -1,34 +1,38 @@
...
 
 ## Overview
 
-At a high level, the CSI Secrets Store driver allows users to create `SecretProviderClass` objects.
+At a high level, the Secrets Store CSI driver allows users to create `SecretProviderClass` objects.
 This object defines which secret provider to use and what secrets to retrieve. When pods requesting CSI volumes
-are created, the CSI Secrets Store driver will send the request to the Vault CSI Provider if the provider
-is `vault`. The Vault CSI Provider will then use Secret Provider Class specified and the pod's service account to retrieve
+are created, the Secrets Store CSI driver will send the request to the Vault Secrets Store CSI Provider if the provider
+is `vault`. The Vault Secrets Store CSI Provider will then use Secret Provider Class specified and the pod's service account to retrieve
```
schavis, line 27:
```suggestion
is `vault`. The Vault Secrets Store CSI provider then uses the specified  `SecretProviderClass` and the service account of the pod to retrieve
```
Style correction: write in active voice, avoid possessives

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
```diff
@@ -1,34 +1,38 @@
...
 the secrets from Vault, and mount them into the pod's CSI volume.
 
 The secret is retrieved from Vault and populated to the CSI secrets store volume during the `ContainerCreation` phase.
 This means that pods will be blocked from starting until the secrets have been read from Vault and written to the volume.
 
 ### Features
 
-The following features are supported by the Vault CSI Provider:
+The following features are supported by the Vault Secrets Store CSI Provider:
```
schavis, line 35:
```suggestion
Vault Secrets Store CSI provider supports the following features:
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/index.mdx
```diff
@@ -57,26 +57,41 @@ There are several ways to try Vault with Kubernetes in different environments.
 
 ### High level comparison of integrations
```
schavis, line 58: This section duplicates the information in the new comparisons page so we should delete it

schavis (changes requested):

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
```diff
@@ -1,34 +1,38 @@
 ---
 layout: docs
-page_title: Vault CSI provider
+page_title: Vault Secrets Store CSI Provider
```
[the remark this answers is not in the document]
digivava: I'm a bit confused by all the uncapitalizing. These individual words aren't proper nouns, no, but they're the names of our offerings, just like an "agent" isn't a proper noun but we call it Vault Agent because that's the product name.

Kubernetes does not capitalize an individual CSI driver, because it's just a category of products. But the Secrets Store CSI [Driver](https://secrets-store-csi-driver.sigs.k8s.io/) capitalizes it, because it's a specific, individual CSI driver product. In the same way as "zoo" is a noun but "Chicago Zoo" takes that noun and makes it part of the proper noun, my assumption is that we would do the same here with Vault Secrets Operator CSI Driver, Vault Secrets Store CSI Provider, etc.

Maybe this is a question for Product? @stevealmyHC Would you prefer the integration names being capitalized, or are you OK lower-casing them like is being suggested?

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/examples.mdx
```diff
@@ -1,14 +1,14 @@
...
 ---
 
-# Vault CSI provider examples
+# Vault Secrets Store CSI Provider examples
 
-The following examples demonstrate how the Vault CSI Provider can be used.
+The following examples demonstrate how the Vault Secrets Store CSI Provider can be used.
 
-~> A common mistake is to not install the CSI Secret Store Driver before using the Vault CSI Provider.
+~> A common mistake is to not install the CSI Secret Store Driver before using the Vault Secrets Store CSI Provider.
```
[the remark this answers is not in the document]
digivava: Yeah it looks like it already is in the prerequisites section of the installation page, I'll just remove it

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/vso-csi.mdx
```diff
@@ -0,0 +1,215 @@
...
+* Secret IDs for the AppRole auth method
+
+## Setup
+
+Installing the Vault Secrets Operator Helm chart with the `"csi.enabled=true"` flag deploys the CSI driver as a DaemonSet running on every node.
+
+```shell-session
+$ helm install --version 0.10.0 --create-namespace --namespace vault-secrets-operator --set "csi.enabled=true" vault-secrets-operator hashicorp/vault-secrets-operator
+```
+
```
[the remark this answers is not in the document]
digivava: You can't create them without running the Helm step first. The Helm chart includes installation of the "Custom Resource Definitions" for those CRs. Without them, Kubernetes won't know what VaultAuth, CSISecrets, etc. mean.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/vso-csi.mdx
```diff
@@ -0,0 +1,215 @@
...
+kind: VaultConnection
+metadata:
+  name: default
+  namespace: vault-secrets-operator-system
+spec:
+  address: http://vault.vault.svc.cluster.local:8200
+  skipTLSVerify: false
+```
+
+See [Vault authentication in detail](/vault/docs/deploy/kubernetes/vso/sources/vault/auth) for more detailed examples of how to set up these resources.
```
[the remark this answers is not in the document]
digivava: We haven't linked to this particular page anywhere above yet, and I've found that this page is the real goldmine of "how to actually get all the complicated auth pieces in VSO working well together", so I would hate to leave it out for users, as I find myself referencing it constantly. (There are a lot of different possible auth workflows, and that page covers various choices.)

Would you prefer I link to it at the end of the previous paragraph, just after the links to the individual API references for each resource?

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx
```diff
@@ -0,0 +1,251 @@
...
+    namespacePatterns:
+      - "default"
+    podNamePatterns:
+      - "^my-app-"
+  syncConfig:
+    containerState:
+      namePattern: "^(app|sidecar)$"
+```
+
+Refer to the full [CSISecrets CR API reference](/vault/docs/deploy/kubernetes/vso/api-reference#csisecrets) for additional configuration options.
```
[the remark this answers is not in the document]
digivava: That subPath field isn't declared on the CSISecrets resource yaml though; it's declared on the Deployment resource yaml. I worry that putting it here would make people think that this field (which is provided by Kubernetes, but isn't necessarily common knowledge) is something they can set on the CSISecrets CR, when it's not.

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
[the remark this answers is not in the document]
digivava: @tvoran Would you be able to take a glance at this `comparisons.mdx` page for accuracy, with your knowledge of Injector and CSI Provider? (Feel free to use the vercel preview on the PR to view it in a more visually discernible way) 

A lot of these were populated from older comparison matrices written by other people, so I am not sure how accurate/confusing they are.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx
```diff
@@ -0,0 +1,251 @@
+---
+layout: docs
+page_title: Installing the Vault Secrets Operator CSI driver
```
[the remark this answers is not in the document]
digivava: The style guide explicitly told me not to use the acronym "VSO" so I've been avoiding it.... is the overall point here just to rearrange it to say "Vault Secrets Operator CSI driver setup" instead?

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx
```diff
@@ -0,0 +1,251 @@
...
+```
+vault write auth/kubernetes/role/example-auth-role \
+    bound_service_account_names=default \
+    bound_service_account_namespaces=default,vault-secrets-operator-system \
+    token_period=120 \
+    token_policies=csi-driver-policy \
+    audience=vault
+```
+
+### Auth resources
```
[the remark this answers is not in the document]
digivava: Edited the above just so I can add it to the batch commit -- I think the "auth" part is required to differentiate it from the other custom resource in Step 5 (the CSISecrets custom resource)

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx
```diff
@@ -0,0 +1,251 @@
...
+(`app_role_0_wrap_info.json`):
+	
+```yaml
+volumeMounts:
+- name: csi-secrets
+  mountPath: /var/run/app_role_0_wrap_info.json
+  subPath: app_role_0_wrap_info.json
+```
+
+</Tip>
```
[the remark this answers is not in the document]
digivava: see other comment about subPath

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx
```diff
@@ -0,0 +1,281 @@
...
+          mountPath: /var/run/csi-secrets
+      volumes:
+      - name: csi-secrets
+        csi:
+          driver: csi.vso.hashicorp.com
+          volumeAttributes:
+            csiSecretsName: my-app-secrets
+            csiSecretsNamespace: vault-secrets-operator-system
+```
+
```
schavis, line 260:
````suggestion
<Tip>

You can also use the Kubernetes
[`subPath`](https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath)
field and the associated file index to limit which secret files a container can
access. For example, the following `subPath` limits access to the first `approle`
secret (`app_role_0_wrap_info.json`):
	
```yaml
volumeMounts:
- name: csi-secrets
  mountPath: /var/run/app_role_0_wrap_info.json
  subPath: app_role_0_wrap_info.json
```

</Tip>

Refer to the full
[CSISecrets CR API reference](/vault/docs/deploy/kubernetes/vso/api-reference#csisecrets)
for additional configuration options.

Next, push the new configuration so the CSI driver pod on your application pod
node responds to the request. If the update succeeds, the `kubelet` publishes
a volume containing all the secrets from the `CSISecrets` resource.
````
Moved from above

File: content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx [hunk 30]
```diff
@@ -0,0 +1,281 @@
...
+- Add the new `CSISecrets` resource to the `volumeMounts` field
+- Ensure that all pod names, namespaces, etc. match the patterns defined in the 
+`accessControl` stanza of your `CSISecrets` resource, and that these pods should have access to the secret data.
+
+Then push the new configuration so the CSI driver pod on your application pod
+node responds to the request. If the update succeeds, the `kubelet` publishes
+a volume containing all the secrets from the `CSISecrets` resource.
+
+For example:
```
schavis, line 221:
```suggestion
For example:

```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx
```diff
@@ -0,0 +1,251 @@
...
+description: >-
+  Install the Vault Secrets Operator CSI driver and 
+  mount secrets into your applications as ephemeral volumes.
+---
+
+# Setup
+
+@include 'alerts/enterprise-only.mdx'
+
+Install the Vault Secrets Operator CSI driver and create a `CSISecrets` resource to allow authorized pods to mount ephemeral volumes containing secrets from Vault.
```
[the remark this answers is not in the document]
schavis: It's not TMI. A how-to guide should provide a summary of what will happen if folks follow the instructions. In this case, that means they'll end up with a set of indexed files full of secrets.

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,142 @@
...
+	increase. Integrations that create independent Vault connections for each pod do
+	not scale well as the number of pods increase.
+	
+	- **Vault Secrets Operator** - shares secret data.
+	- **Vault Secrets Store CSI provider** - creates independent Vault connections.
+	- **Vault Agent Injector** - creates independent Vault connections.
+	
+	</Tab>
+	
+	</Tabs>
```
schavis, line 120:
```suggestion
<Tabs>
	
<Tab heading="Load on Vault">
	
- **Vault Secrets Operator** - Lowest load due to per-CRD connections and
  cluster-local caching.
- **Vault Secrets Store CSI provider** - Higher load due to per-pod connections.
- **Vault Agent Injector** - Highest load due to per-pod connections and sidecar
  pattern.
	
</Tab>
	
<Tab heading="Secret availability">
	
Integrations that use ephemeral storage have variable availability because
secret access depends on Vault availability.
	
- **Vault Secrets Store CSI provider** - Variable availability.
- **Vault Agent Injector** - Variable availability.
- **Vault Secrets Operator** - Best availability. Kubernetes Secrets act
  as durable cache.
	
</Tab>
	
<Tab heading="I/O, CPU, and memory consumption">
	
- **Vault Secrets Operator** - Low consumption.
- **Vault Secrets Store CSI provider** - Medium consumption.
- **Vault Agent Injector** - High consumption due to the number of sidecar containers
  per pod.
	
</Tab>
	
<Tab heading="Scalability">

Integrations that share secret data across pods scale well as the number of pods
increase. Integrations that create independent Vault connections for each pod do
not scale well as the number of pods increase.
	
- **Vault Secrets Operator** - shares secret data.
- **Vault Secrets Store CSI provider** - creates independent Vault connections.
- **Vault Agent Injector** - creates independent Vault connections.
	
</Tab>
	
</Tabs>
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,142 @@
...
+secret access depends on Vault availability.
+	
+- **Vault Secrets Store CSI provider** - Variable availability.
+- **Vault Agent Injector** - Variable availability.
+- **Vault Secrets Operator** - Best availability. Kubernetes Secrets act
+  as durable cache.
+	
+</Tab>
+	
+<Tab heading="I/O, CPU, and memory consumption">
```
tvoran, line 96: Maybe just "Resource consumption"?

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,142 @@
...
+	
+- **Vault Secrets Operator** - Low consumption.
+- **Vault Secrets Store CSI provider** - Medium consumption.
```
tvoran, line 99: What are the reasons for the difference in consumption for these two?

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,138 @@
...
+- **Vault Agent Injector** - Variable availability.
+- **Vault Secrets Operator** - Best availability. Kubernetes Secrets act as durable cache.
+	
+</Tab>
+	
+<Tab heading="I/O, CPU, and memory consumption">
+	
+- **Vault Secrets Operator** - Low consumption.
+- **Vault Secrets Store CSI provider** - Medium consumption.
+- **Vault Agent Injector** - High consumption due to the number of sidecar containers per pod.
```
tvoran, line 100: It's just one sidecar and/or one init container per Pod, which aren't running simultaneously; maybe just drop the "per pod" bit?

Side note: Do we have data showing high cpu and memory usage for Agent? The default CPU and memory resource requests are pretty high, but they can be tuned down fairly low IIRC.

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,138 @@
...
+</Tab>
+	
+<Tab heading="Secret availability">
+
+Integrations that use ephemeral storage have variable availability because
+secret access depends on Vault availability.
+
+- **Vault Secrets Store CSI provider** - Variable availability.
+- **Vault Agent Injector** - Variable availability.
+- **Vault Secrets Operator** - Best availability. Kubernetes Secrets act as durable cache.
```
tvoran, line 92: > secret access depends on Vault availability.

Isn't that true for any of these? i.e. nothing can fetch secrets if Vault is down. If a secret is fetched from Vault and rendered, Vault subsequently going down isn't going to clear the rendered secret for CSI provider or Agent, is it?

Integrations using ephemeral storage means the rendered secret lifecycle is tied to the Pod lifecycle, instead of being independent for VSO, maybe re-frame along those lines?

schavis (approved):

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,136 @@
...
+
+Use the information below to compare the available integrations and choose the right solution for your needs.
+
+## Features at a glance
+
+| Feature                             | Vault Secrets Operator                                             | Vault Secrets Store CSI provider   | Vault Agent Injector
+	| ----------------------------------- | --------------------------------- | ---------------------------- | --------------
+	| Supported secret types              | Static/KV, PKI, Dynamic, Database, AppRole secret IDs            | All                  | All
+	| Supported auth methods              | K8s, AppRole, GCP, AWS, JWT                                      | K8s, JWT             | K8s, other [auto-auth methods](/vault/docs/agent-and-proxy/autoauth)
+	| Storage model                       | K8s Secrets or ephemeral volumes with CSI driver                 | Ephemeral volumes    | Ephemeral volumes
```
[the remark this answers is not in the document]
digivava: Thanks for catching this here; I just learned this myself!

tvoran (approved):
One more small thing caught my eye on the comparisons page, but 👍

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,54 @@
+---
+layout: docs
+page_title: Direct secrets
```
benashz, line 3: Is this the new name?

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,54 @@
...
+  Mount secrets from Vault directly into application containers with the Vault
+  Secrets Operator CSI driver.
+---
+
+# Direct secrets
+
+@include 'alerts/enterprise-only.mdx'
+
+While the Vault Secrets Operator lets you store Vault secrets in clusters as native
+Kubernetes Secrets, you can also use the "direct secrets" feature to mount Vault secrets directly to containers as volumes.
```
benashz, line 14: might be good to link out the to the K8s `Volumes` docs somewhere.

benashz, line 14: Can we explicitly call out that this new feature relies on the Kubernetes CSI and is implemented a CSI driver?

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,54 @@
...
+The Vault Secrets Operator CSI driver supports:
+- Syncing multiple secrets from multiple plugins. Supported plugins include:
+  - [Key/Value v1](/vault/docs/secrets/kv/kv-v1)
+  - [Key/Value v2](/vault/docs/secrets/kv/kv-v2)
+- Automatic generation and mounting of response-wrapped AppRole secret IDs as a
+  [trusted orchestrator](/vault/tutorials/app-integration/secure-introduction?in=vault%2Fapp-integration#trusted-orchestrator) for authentication between application pods and Vault.
+- Helm-based [installation](/vault/docs/platform/k8s/vso/installation).
+- [Secret data transformation](/vault/docs/platform/k8s/vso/secret-transformation).
+
+The Vault Secrets Operator CSI Driver currently does not support the [instant updates](/vault/docs/deploy/kubernetes/vso/sources/vault/instant-updates) feature.
```
benashz, line 29: Worth calling out that generically KV secret updates are provisioned on as new Pods come online, or whenever a matching container crashes/exists

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,54 @@
...
+- Helm-based [installation](/vault/docs/platform/k8s/vso/installation).
+- [Secret data transformation](/vault/docs/platform/k8s/vso/secret-transformation).
+
+The Vault Secrets Operator CSI Driver currently does not support the [instant updates](/vault/docs/deploy/kubernetes/vso/sources/vault/instant-updates) feature. 
+Application containers must self-restart to receive updated secret information after a secret is updated in Vault.
+
+## Secrets flow: Traditional VSO vs CSI driver
+
+In the traditional workflow, the Vault Secrets Operator caches Vault secrets as Kubernetes Secrets, which are stored in etcd. 
+The Vault Secrets Operator watches for changes to custom resources such as [VaultStaticSecret](/vault/docs/platform/k8s/vso/api-reference#vaultstaticsecret) or [VaultDynamicSecret](/vault/docs/platform/k8s/vso/api-reference#vaultdynamicsecret).
```
benashz, line 35: Can you use/extend the common-links.mdx for references to the VSO API.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,54 @@
...
+When the resource definition changes, Vault Secrets Operator overwrites secret data in the associated Kubernetes Secret with the new information.
+
+[![Diagram showing a secret flowing from Vault, through VSO, into the associated Kubernetes Secrets objects across all nodes in an app namespace](/img/vso_secrets_flow.png)](/img/vso_secrets_flow.png)
+
+In the CSI driver workflow, the driver does not cache secrets as Kubernetes Secrets, so there is no storage in `etcd`.
+The driver watches for changes to a custom resource (CR) called `CSISecrets`. The
+[CSISecrets](/vault/docs/deploy/kubernetes/vso/api-reference#csisecrets) resource
+contains a declarative list of secrets to fetch and an access control
+configuration that only allows the designated pods to mount those secrets.
+When the spec of an approved pod references the `CSISecrets` resource as a volume to mount, the Vault Secrets Operator CSI driver populates
```
benashz, line 45: typically Pod is capitalized.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,54 @@
...
+In the CSI driver workflow, the driver does not cache secrets as Kubernetes Secrets, so there is no storage in `etcd`.
+The driver watches for changes to a custom resource (CR) called `CSISecrets`. The
+[CSISecrets](/vault/docs/deploy/kubernetes/vso/api-reference#csisecrets) resource
+contains a declarative list of secrets to fetch and an access control
+configuration that only allows the designated pods to mount those secrets.
+When the spec of an approved pod references the `CSISecrets` resource as a volume to mount, the Vault Secrets Operator CSI driver populates 
+that volume with the associated secrets for all referencing containers in the pod.
+
+The pod starts up with all of the secrets from the `CSISecrets` resource mounted
+at the configured path (for example, `/var/run/csi-secrets`) as individual files.
```
benashz, line 49: It might be good to call out how these volumes are configured.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx
```diff
@@ -0,0 +1,290 @@
...
+
+
+## Step 1: Install driver
+
+Install the Vault Secrets Operator Helm chart with the `csi.enabled` flag set to `true` to deploy the CSI
+driver as a DaemonSet running on every node:
+
+```shell-session
+$ helm install                         \
+    --version 0.10.0                   \
```
benashz, line 37: We will need to bump this to be 1.0.0 before releasing these docs.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/installation.mdx
```diff
@@ -41,6 +41,25 @@ Then install the Operator:
...
 ```
 
+## CSI driver option
+
+You can also use the Vault Secrets Operator Helm chart to deploy a DaemonSet of the
+[Vault Secrets Operator CSI driver](/vault/docs/platform/k8s/vso/csi) with
+the `"csi.enabled"` flag:
+
+```shell-session
+$ helm install                          \
```
benashz, line 51: We already have dedicated installation docs for both new installations and upgrades. Would it make sense to link to those from here?

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,136 @@
...
+	| ----------------------------------- | --------------------------------- | ---------------------------- | --------------
+	| Supported secret types              | Static/KV, PKI, Dynamic, Database, AppRole secret IDs            | All                  | All
+	| Supported auth methods              | K8s, AppRole, GCP, AWS, JWT                                      | K8s, JWT             | K8s, other [auto-auth methods](/vault/docs/agent-and-proxy/autoauth)
+	| Storage model                       | K8s Secrets or ephemeral volumes with CSI driver                 | Temporary K8s Secrets or ephemeral volumes    | Ephemeral volumes
+	| Kubernetes-native                   | Yes                                                              | Yes                  | No
+	| Secret data persistence             | Persisted in `etcd` or ephemeral with CSI driver                 | Temporarily persisted or ephemeral | Ephemeral
+	| Secret data templating              | Yes                                                              | No                   | Yes
+	| Requires Vault for Pod autoscaling  | No                                                               | Yes                  | Yes
```
benashz, line 26: Only in the non CSI case.

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,136 @@
...
+	| Secret sharing between applications | Yes                                                              | Yes                  | No
+
+## Vault Secrets Operator
+
+The [Vault Secrets Operator](/vault/docs/platform/k8s/vso) synchronizes secrets from Vault to native Kubernetes Secrets. 
+It operates by watching Custom Resource Definitions (CRDs) and syncing secret data from Vault to Kubernetes Secrets, which applications can consume using standard Kubernetes patterns.
```
benashz, line 32: I would replace watching with reconciling

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,136 @@
...
+
+## Use case recommendations
+
+The appropriate integration for your environment depends on the applicable
+security policies, secret data governance, ease of use requirements, and
+system availability guarantees.
+
+| Limitation                                        | Vault Secrets Operator            | Vault Secrets Store CSI provider　| Vault Agent Injector
+| ------------------------------------------------- | --------------------------------- | ---------------------------- | --------------
+| Requires privileged pod access                    | No, except with CSI driver        | Yes                          | No
```
benashz, line 60: nit: capitalize Pod as it is a resource like CSISecrets, etc.

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,136 @@
...
+| Flexibility in how to consume secrets             | Yes                               | Yes                          | No
+| Low resource consumption (I/O, CPU, Memory)       | Yes                               | Yes                          | No
+| Secret data limited to ephemeral volumes          | No                                | No                           | Yes
+| Pod Autoscaling dependent on Vault availability   | No                                | Yes                          | Yes
```
benashz, line 64: asterix - only in the non CSI case.

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,136 @@
...
+to caching in Kubernetes Secrets, unless using the Vault Secrets Operator CSI driver.
+- **Vault Secrets Store CSI provider** - Secret lifecycle depends on pod lifecycle, even with the 
+option to cache in Kubernetes Secrets (they will be deleted after the pod terminates).
+- **Vault Agent Injector** - Secret lifecycle depends on pod lifecycle.
+
+</Tab>
+	
+<Tab heading="Resource consumption">
+
+- **Vault Secrets Operator** - Lower consumption due to one client process needed per cluster.
```
benashz, line 97: In the case of CSI it is one driver instance per Node

benashz (commented):
Looking really good! I added some feedback for your consideration.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
digivava: I do so in the next line after this, including a link to the k8s CSI docs. After this first "transition" paragraph which I added to help connect the dots for HashiConf keynote viewers, the rest of the docs discuss it as a CSI driver.

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
digivava: Yeah let me add that back in; I'd mentioned that when VSO-CSI was its own bullet point but now we're discussing it as just part of VSO.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/installation.mdx
digivava: I think this is already the installation docs page that you're talking about. This is just a new small section in that doc, explaining the `"csi.enabled" flag, but the rest is all the same as before. (Unless you're talking about a different doc I don't know about)

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
benashz: :+1:

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,57 @@
...
+  Mount secrets from Vault directly into application containers with the Vault
+  Secrets Operator CSI driver.
+---
+
+# Direct secrets
+
+@include 'alerts/enterprise-only.mdx'
+
+While the Vault Secrets Operator lets you store Vault secrets in clusters as native
+Kubernetes Secrets, you can also use the "direct secrets" feature to mount Vault secrets directly to containers as [volumes](https://kubernetes.io/docs/concepts/storage/volumes/).
```
benashz, line 14: nit: can we call these "ephemeral volumes" throughout. They are defined as such in the `CSIDriver` resource that ships with the chart.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+The direct secrets feature installs the Vault Secrets
+Operator [CSI](https://github.com/container-storage-interface/spec/blob/master/spec.md) driver, which hooks into the Pod volume lifecycle directly,
+bypassing Kubernetes Secrets.
```
benashz, line 18: I wonder if we could replace the "bypass" part with something like "secret data from Vault is stored in ephemeral Kubernetes Volumes" or something to that effect?

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,136 @@
...
+</Tab>
+	
+</Tabs>
+
+
+## Migration requirements 
+
+- Migrating **to** Vault Secrets Operator requires:
+    1. Creating CRDs for each relevant secret type. For example, `VaultStaticSecret`, `VaultDynamicSecret`, `VaultPKISecret`, or `CSISecrets`.
+    1. Updating applications to use Kubernetes Secrets.
```
benashz, line 124: Would it be worth linking out to the API docs here?

digivava:
@schavis The team has completed their review and I have addressed all their comments (including comments off of GitHub from Product). So this should be ready for re-review whenever you're able.

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,140 @@
...
+The [Vault Secrets Operator](/vault/docs/platform/k8s/vso) synchronizes secrets from Vault to native Kubernetes Secrets. 
+It operates by reconciling Custom Resource Definitions (CRDs) and syncing secret data from Vault to Kubernetes Secrets, which applications can consume using standard Kubernetes patterns.
+Vault Enterprise users can also configure it to run a [CSI driver](/vault/docs/platform/k8s/vso/csi), which mounts secrets directly from Vault to application containers as volumes, bypassing Kubernetes Secrets. 
+
+**Best for:** Teams who prefer native Kubernetes workflows and want the option of either durable secret storage within the cluster, or written to ephemeral volumes.
+
+## Vault Secrets Store CSI provider
+
+The [Vault Secrets Store CSI provider](/vault/docs/platform/k8s/csi) uses the vendor-neutral [Secrets Store CSI driver](https://secrets-store-csi-driver.sigs.k8s.io/) to mount secrets from Vault as ephemeral volumes. 
+Vault Agent can be deployed alongside the provider to manage [caching and renewals](/vault/docs/agent-and-proxy/agent/caching#caching-and-renewals).
```
schavis, line 40:
```suggestion
You can deploy Vault Agent alongside the provider to manage [caching and renewals](/vault/docs/agent-and-proxy/agent/caching#caching-and-renewals).
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+**Best for:** Organizations using multiple secret stores in addition to Vault, or those who prefer vendor-neutral CSI standards.
+
+## Vault Agent Injector
+
+The [Vault Agent Injector](/vault/docs/platform/k8s/injector) injects Vault Agent sidecar containers into Pods. 
+The Agent containers authenticate with Vault and render secrets to a shared memory volume for consumption by application containers.
+
+**Best for:** Applications that want to reference multiple Vault secrets in one template, or need a wider array of auth methods.
```
schavis, line 49:
```suggestion
**Best for:** Applications that want to reference multiple Vault secrets in one template or need a wider array of auth methods.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,140 @@
...
+system availability guarantees.
+
+| Limitation                                        | Vault Secrets Operator            | Vault Secrets Store CSI provider　| Vault Agent Injector
+| ------------------------------------------------- | --------------------------------- | ---------------------------- | --------------
+| Requires privileged Pod access                    | No, except with CSI driver        | Yes                          | No
+| Flexibility in how to consume secrets             | Yes                               | Yes                          | No
+| Low resource consumption (I/O, CPU, Memory)       | Yes                               | Yes                          | No
+| Secret data limited to ephemeral volumes          | No                                | No                           | Yes
+| Pod Autoscaling dependent on Vault availability   | No, except with CSI driver        | Yes                          | Yes
+| Personas easily separated via K8s [RBAC] policies | Yes                               | No                           | No
```
schavis, line 65:
```suggestion
| Limitation                                        | Vault Secrets Operator              | Vault Secrets Store CSI provider　| Vault Agent Injector
| ------------------------------------------------- | ----------------------------------- | ---------------------------- | --------------
| Requires privileged Pod access                    | No (default) <br/> Yes (CSI driver) | Yes                          | No
| Flexibility in how to consume secrets             | Yes                                 | Yes                          | No
| Low resource consumption (I/O, CPU, Memory)       | Yes                                 | Yes                          | No
| Secret data limited to ephemeral volumes          | No                                  | No                           | Yes
| Pod Autoscaling dependent on Vault availability   | No (default) <br/> Yes (CSI driver) | Yes                          | Yes
| Personas easily separated via K8s [RBAC] policies | Yes                                 | No                           | No
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,140 @@
...
+- **Vault Secrets Operator** - Lowest load due to per-CRD connections and cluster-local caching. 
+Single manager for an entire cluster. If using the CSI driver, there will be one Vault connection pool per Node.
+- **Vault Secrets Store CSI provider** - Higher load due to per-Pod connections.
+- **Vault Agent Injector** - Highest load due to per-Pod connections and sidecar pattern.
+	
+</Tab>
+	
+<Tab heading="Rendered secret lifecycle">
+
+- **Vault Secrets Operator** - Secret lifecycle is independent of Pod lifecycle due
```
schavis, line 88: Break this into two lines

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,140 @@
...
+to caching in Kubernetes Secrets, unless using the Vault Secrets Operator CSI driver.
+- **Vault Secrets Store CSI provider** - Secret lifecycle depends on Pod lifecycle, even with the 
+option to cache in Kubernetes Secrets (they will be deleted after the Pod terminates).
+- **Vault Agent Injector** - Secret lifecycle depends on Pod lifecycle.
+
+</Tab>
+	
+<Tab heading="Resource consumption">
+
+- **Vault Secrets Operator** - Lowest consumption due to one client process needed per Kubernetes cluster.
```
schavis, line 98: Break this into two lines

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,140 @@
...
+- **Vault Agent Injector** - Highest load due to per-Pod connections and sidecar pattern.
+	
+</Tab>
+	
+<Tab heading="Rendered secret lifecycle">
+
+- **Vault Secrets Operator** - Secret lifecycle is independent of Pod lifecycle due 
+to caching in Kubernetes Secrets, unless using the Vault Secrets Operator CSI driver.
+- **Vault Secrets Store CSI provider** - Secret lifecycle depends on Pod lifecycle, even with the 
+option to cache in Kubernetes Secrets (they will be deleted after the Pod terminates).
```
schavis, line 91:
```suggestion
- **Vault Secrets Store CSI provider** - Secret lifecycle depends on Pod lifecycle, even with the 
option to cache in Kubernetes Secrets (they will be deleted after the Pod terminates).```suggestion
- **Vault Secrets Store CSI provider** - Secret lifecycle depends on Pod
  lifecycle as Vault deletes cached Kubernetes Secrets when the Pod terminates.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/comparisons.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+1. Creating CRDs for each relevant secret type, such as `VaultStaticSecret`, `VaultDynamicSecret`, `VaultPKISecret`, or `CSISecrets`. Refer to the [API reference](/vault/docs/platform/k8s/vso/api-reference) for configuration details.
+Note that the Vault Secrets Operator CSI driver only supports KV secrets and AppRole secret IDs.
+1. Updating applications to use Kubernetes Secrets.
+1. Migrating from Vault Agent Injector may require updating your secret templates to 
+use Vault Secrets Operator's `SecretTransformation` resource. 
+While both tools use Go templating, Vault Agent Injector templates may not work with the `SecretTransformation` resource out of the box.
+1. If using the [direct secrets](/vault/docs/platform/k8s/vso/csi) functionality, you will also need:
+  1. A Vault Enterprise server.
+  1. Updated Pod specifications that reference CSI volumes.
```
schavis, line 132:
```suggestion
1. To use [direct secrets](/vault/docs/platform/k8s/vso/csi) you also need:
  1. A Vault Enterprise license.
  1. Updated Pod specifications that reference CSI volumes.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
```diff
@@ -1,34 +1,38 @@
...
-The Vault CSI Provider allows pods to consume Vault secrets using
-[CSI Secrets Store](https://github.com/kubernetes-sigs/secrets-store-csi-driver) volumes.
+The Vault Secrets Store CSI provider allows pods to consume Vault secrets using
+[Secrets Store CSI](https://github.com/kubernetes-sigs/secrets-store-csi-driver) volumes. 
+Alternatively, you can deploy HashiCorp's
+[Vault Secrets Operator CSI driver](/vault/docs/platform/k8s/vso/csi) as part of your
+Vault Secrets Operator installation.
 
-The Vault CSI provider is a fully supported component of HashiCorp Vault.
+The Vault Secrets Store CSI provider is a fully supported component of HashiCorp Vault.
```
schavis, line 17:
```suggestion
```
I know this was part of the original doc, but it's extraneous. We don't document anything in our product docs that *isn't* fully supported

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
```diff
@@ -1,34 +1,38 @@
...
 
-~> The Vault CSI Provider requires the [CSI Secret Store](https://github.com/kubernetes-sigs/secrets-store-csi-driver)
+~> The Vault Secrets Store CSI provider requires the [CSI Secret Store](https://github.com/kubernetes-sigs/secrets-store-csi-driver)
 Driver to be installed.
 
 ## Overview
 
-At a high level, the CSI Secrets Store driver allows users to create `SecretProviderClass` objects.
+At a high level, the Secrets Store CSI driver allows users to create `SecretProviderClass` objects.
```
schavis, line 24:
```suggestion
At a high level, the Secrets Store CSI driver lets you create `SecretProviderClass` objects.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/csi/index.mdx
```diff
@@ -1,34 +1,38 @@
...
 This object defines which secret provider to use and what secrets to retrieve. When pods requesting CSI volumes
-are created, the CSI Secrets Store driver will send the request to the Vault CSI Provider if the provider
-is `vault`. The Vault CSI Provider will then use Secret Provider Class specified and the pod's service account to retrieve
+are created, the Secrets Store CSI driver will send the request to the Vault Secrets Store CSI provider if the provider
+is `vault`. The Vault Secrets Store CSI provider then uses the specified `SecretProviderClass` and the service account of the pod to retrieve
```
schavis, line 27:
```suggestion
are created, the Secrets Store CSI driver sends the request to the Vault Secrets Store CSI provider if the provider
is `vault`. The Vault Secrets Store CSI provider then uses the specified `SecretProviderClass` and the service account of the pod to retrieve
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,59 @@
+---
+layout: docs
+page_title: Direct secrets
```
schavis, line 3:
```suggestion
page_title: Use direct secrets in Kubernetes
```
Style correction: docs titles should not be features names. Instead, they should use direct language to indicate whether the information is conceptual or instructional.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,59 @@
...
+description: >-
+  Mount secrets from Vault directly into application containers with the Vault
+  Secrets Operator CSI driver.
+---
+
+# Direct secrets
```
schavis, line 9:
```suggestion
# Use direct secrets in Kubernetes with CSI
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+@include 'alerts/enterprise-only.mdx'
+
+The direct secrets feature is in **public beta**.
```
schavis:
```suggestion
@include 'alerts/beta.mdx'
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+While the Vault Secrets Operator lets you store Vault secrets in clusters as native
+Kubernetes Secrets, you can also use the "direct secrets" feature to mount Vault secrets directly to containers as Kubernetes [ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/).
+
+The direct secrets feature installs the Vault Secrets
+Operator [CSI](https://github.com/container-storage-interface/spec/blob/master/spec.md) driver. 
+The CSI driver hooks into the Pod volume lifecycle, allowing applications to store secret data from Vault in ephemeral volumes with the same lifecycle as the application Pod itself.
```
schavis, line 18:
```suggestion
While the Vault Secrets Operator lets you store Vault secrets in clusters as
native Kubernetes Secrets, you must install the Vault Secrets Operator
[CSI driver](https://github.com/container-storage-interface/spec/blob/master/spec.md)
to use "direct secrets" functionality. 

The CSI driver hooks into the Pod volume lifecycle so applications can store
secret data from Vault in
[ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/)
with the same lifecycle as the application Pod itself.
```

Style correction: write in active voice, focus on the job to be done and the tools involved not features

Unless there's a button in Vault that installs and deploys the driver for you 
(which is not what the previous docs implied), the feature does not install
anything. We're just calling the result of deploying the driver "direct secrets".

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
```diff
@@ -0,0 +1,59 @@
...
+The Vault Secrets Operator CSI driver supports:
+- Syncing multiple secrets from multiple plugins. Supported plugins include:
+  - [Key/Value v1](/vault/docs/secrets/kv/kv-v1)
+  - [Key/Value v2](/vault/docs/secrets/kv/kv-v2)
+- Automatic generation and mounting of response-wrapped AppRole secret IDs as a
+  [trusted orchestrator](/vault/tutorials/app-integration/secure-introduction?in=vault%2Fapp-integration#trusted-orchestrator) for authentication between application Pods and Vault.
+- Helm-based [installation](/vault/docs/platform/k8s/vso/installation).
+- [Secret data transformation](/vault/docs/platform/k8s/vso/secret-transformation).
+
+The Vault Secrets Operator CSI driver does not support the [instant updates](/vault/docs/deploy/kubernetes/vso/sources/vault/instant-updates) feature.
```
schavis, line 29:
```suggestion
The Vault Secrets Operator CSI driver does not support [instant updates for `VaultStaticSecrets`](/vault/docs/deploy/kubernetes/vso/sources/vault/instant-updates).
```

schavis (changes requested):

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/helm.mdx
```diff
@@ -726,6 +751,160 @@ Use these links to navigate to a particular top-level stanza.
...
+    - `backoffOnSecretSourceError` ((#v-csi-driver-backoffonsecretsourceerror)) (`object: ""`) - Backoff settings for the CSI driver. These settings control the backoff behavior
+      when the driver encounters an error while fetching secrets from the SecretSource.
+      For example given the following settings:
+        initialInterval: 5s
+        maxInterval: 60s
+        randomizationFactor: 0.5
+        multiplier: 1.5
+
+      The backoff retry sequence might be something like:
+       5.5s, 7.5s, 11.25s, 16.87s, 25.3125s, 37.96s, 56.95, 60.95s...
```
schavis, line 854:
````suggestion
    - `backoffOnSecretSourceError` ((#v-csi-driver-backoffonsecretsourceerror))
      (`object: ""`) - YAML object describing backoff behavior for the CSI driver
      when the  driver encounters an error fetching secrets from `SecretSource`.
      For example, the following configuration creates a backoff retry sequence
      like 5.5s, 7.5s, 11.25s, 16.87s, 25.3125s, 37.96s, 56.95, 60.95s, etc.:
      ```yaml
      "initialInterval": 5s
      "maxInterval": 60s
      "randomizationFactor": 0.5
      "multiplier": 1.5
      ```
````

File: content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/helm.mdx [hunk 70]
```diff
@@ -726,6 +751,160 @@ Use these links to navigate to a particular top-level stanza.
...
+
+      - `initialInterval` ((#v-csi-driver-backoffonsecretsourceerror-initialinterval)) (`duration: 5s`) - Initial interval between retries.
+
+      - `maxInterval` ((#v-csi-driver-backoffonsecretsourceerror-maxinterval)) (`duration: 60s`) - Maximum interval between retries.
+
+      - `maxElapsedTime` ((#v-csi-driver-backoffonsecretsourceerror-maxelapsedtime)) (`duration: 0s`) - Maximum elapsed time without a successful sync from the secret's source.
+        It's important to note that setting this option to anything other than
+        its default will result in the secret sync no longer being retried after
+        reaching the max elapsed time.
```
schavis, line 863:
````suggestion
      - `maxElapsedTime` ((#v-csi-driver-backoffonsecretsourceerror-maxelapsedtime))
        (`duration: 0s`) - Maximum elapsed time without a successful sync from
        the secret source. Setting an explicit `maxElapsedTime` tells VSO not to
        retry the secret sync once it reaches the max elapsed time.
````

File: content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/helm.mdx [hunk 71]
```diff
@@ -726,6 +751,160 @@ Use these links to navigate to a particular top-level stanza.
...
+
+      - `randomizationFactor` ((#v-csi-driver-backoffonsecretsourceerror-randomizationfactor)) (`float: 0.5`) - Randomization factor randomizes the backoff interval between retries.
+        This helps to spread out the retries to avoid a thundering herd.
+        If the value is 0, then the backoff interval will not be randomized.
+        It is recommended to set this to a value that is greater than 0.
```
schavis, line 868:
```suggestion
      - `randomizationFactor` ((#v-csi-driver-backoffonsecretsourceerror-randomizationfactor))
        (`float: 0.5`) - Custom factor used to randomize the backoff interval
        between retries, spread out the retries, and avoid a thundering herd.
        We strongly recommend using a randomization factor, but you can set
        `randomizationFactor` to `0` to disable randomization
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/helm.mdx
```diff
@@ -726,6 +751,160 @@ Use these links to navigate to a particular top-level stanza.
...
+
+      - `multiplier` ((#v-csi-driver-backoffonsecretsourceerror-multiplier)) (`float: 1.5`) - Sets the multiplier that is used to increase the backoff interval between retries.
+        This value should always be set to a value greater than 0.
+        The value must be greater than zero.
```
schavis, line 872:
```suggestion
      - `multiplier` ((#v-csi-driver-backoffonsecretsourceerror-multiplier))
        (`float: 1.5`) - Sets a multiplier used to increase the backoff
        interval between retries. You cannot set `multiplier` a value less than
        or equal to `0`.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/helm.mdx
```diff
@@ -726,6 +751,160 @@ Use these links to navigate to a particular top-level stanza.
...
+
+  - `livenessProbe` ((#v-csi-livenessprobe))
+
+    - `image` ((#v-csi-livenessprobe-image)) - Image sets the repo and tag of the image to use for the liveness probe.
```
schavis, line 876:
```suggestion
    - `image` ((#v-csi-livenessprobe-image)) - The repo and tag of the image to
      use for the liveness probe.
```

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
digivava: VSO-CSI doesn't use the custom resource called VaultStaticSecrets so I'm just going to put Instant Updates here with the link, I think

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/index.mdx
digivava: Agreed. I've modified this to match what you're saying while also trying to keep the word "mount directly to containers" in there which I think Product cares about, to provide a definition for what "direct secrets" actually means. Hopefully what I've got there now satisfies both.

content/vault/v1.20.x/content/docs/deploy/kubernetes/vso/csi/setup.mdx
```diff
@@ -0,0 +1,286 @@
...
+
+You must create a
+[`VaultAuth`](/vault/docs/deploy/kubernetes/vso/api-reference#vaultauth) and
+[`VaultConnection`](/vault/docs/deploy/kubernetes/vso/api-reference#vaultconnection), 
+as defined in the Custom Resource Definitions (CRDs) that were deployed by the Vault Secrets Operator Helm chart, 
+so the CSI driver can connect to Vault. 
+
+If desired, you can reference existing auth resources from Vault Secrets Operator.
+
+Create a YAML file with these resources, and apply it to your cluster. For example:
```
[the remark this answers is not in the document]
digivava: I happen to be using a single YAML file here, but some people might create each resource in multiple YAML files. The file is not important; in fact some people create resources via json, maybe even by piping their yaml/json directly to the kubectl CLI. And even then, while most people use the `kubectl` CLI to create resources, others might use some kind of UI or other API client to do it. 

Because of this, there is a base assumption for practitioners reading these docs that they know how to create Kubernetes resources given some YAML (just like the k8s docs). If this were on the Tutorials site, maybe it'd make sense to prescribe the tool/input format to use.

So I've switched to active voice but changed this to be more about creating the resource. Hope that makes sense!

schavis (approved):

PR #714, opened 2025-08-11
Updating terraform mcp server with new config options

gautambaghel:


trujillo-adam:
@gautambaghel - I decided to move the reference information to a separate page. Now, the overview and deploy pages are much cleaner. I also streamlined the prompt page. LMK if we are OK to merge or if there is anything else to address.

PR #749, opened 2025-08-13
Update descriptions of `-state` and `-state-out` flags to be explicit that they only work with the local backend

SarahFrench:
Fixes https://github.com/hashicorp/terraform/issues/29470

Looking through the documentation there are lots of places where the `-state` and `-state-out` flags are warned to be legacy options that only work with the local backend. For example [this state refactoring guidance page has a red warning](https://developer.hashicorp.com/terraform/language/state/refactor#move-resources-directly-to-a-new-state-file) about the flags, and [docs for commands use text like this](https://developer.hashicorp.com/terraform/cli/commands/apply#:~:text=For%20configurations%20using%20the%20local%20backend%20only%2C%20terraform%20apply%20also%20accepts%20the%20legacy%20options%20%2Dstate%2C%20%2Dstate%2Dout%2C%20and%20%2Dbackup.):

>For configurations using [the local backend](https://developer.hashicorp.com/terraform/language/backend/local) only, terraform apply also accepts the legacy options [-state, -state-out, and -backup](https://developer.hashicorp.com/terraform/language/backend/local#command-line-arguments).

In this PR I'm trying to find any last mentions of the flags that aren't caveated in a similar way.

I've replaced "Ignored when remote state is used." with "Legacy option for the local backend only." as it's more specific and more correct; a user of the `s3` backend may believe that it'd work with that backend, as it's not the `remote` backend.

### DIFF
--- a/content/terraform/v1.1.x/docs/cli/commands/output.mdx
+++ b/content/terraform/v1.1.x/docs/cli/commands/output.mdx
@@ -30,7 +30,7 @@ The command-line flags are all optional. The list of available flags are:
   for processing complex data types.
 * `-no-color` - If specified, output won't contain any color.
 * `-state=path` - Path to the state file. Defaults to "terraform.tfstate".
-  [Legacy option](/terraform/language/backend/local#command-line-arguments) for the local backend only.
+  [Legacy option for the local backend only](/terraform/language/backend/local#command-line-arguments).
 
 -> **Note:** When using the `-json` or `-raw` command-line flag, any sensitive
 values in Terraform state will be displayed in plain text. For more information,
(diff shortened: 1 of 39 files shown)


SarahFrench:
> I'm happy with the wording but have one question - Since we are linking to the local backend, should the link be anchored to that as well? e.g. `Legacy option for the [local backend]((/terraform/language/backend/local#command-line-arguments)) only`

That's fair - as the link goes to a legacy options section in the local backend page I've made the whole sentence the link. I worry if we linked using just 'local backend' then users might not realise it's a link to a specific section.

radeksimko (approved):

PR #798, opened 2025-08-14
HYOK roles can be configured in AWS with Web Identity

JarrettSpiker:
AWS IAM recently released an update which allows roles to be configured for HYOK in specific using the "Web Identity" entity type.

Remove the note from our docs which warns users against using the Web Identity entity type.

CleanShot 2025-08-14 at 10 33 15@2x


tylerwolf (approved):

rkoron007 (approved):
I always celebrate removing callouts that are no longer necessary! 🎉

PR #819, opened 2025-08-17
Script to check valid redirects

im2nguyen:
ThIs PR creates a helper script that checks for invalid redirects (destination is not in `/api/all-docs-paths` or if it's a potential daisy chain -- an destination is a source for another redirect).

You can test this script using:

```
./scripts/check_redirects.sh > redirect_analysis.txt
```

### DIFF
--- a/scripts/check_redirects.sh
+++ b/scripts/check_redirects.sh
@@ -1,0 +1,184 @@
+#!/bin/bash
+
+# Content API redirect checker - validates redirects and detects daisy chains
+# Usage: ./check_redirects.sh [BASE_URL] [PRODUCTS]
+# Requirements: jq, curl
+
+set -e
+
+# Configuration
+BASE_URL="${BASE_URL:-http://localhost:8080}"
+PRODUCTS="${PRODUCTS:-}"
+
+# Colors for output
+RED='\033[0;31m'
+YELLOW='\033[1;33m'
+GREEN='\033[0;32m'
+BLUE='\033[0;34m'
+NC='\033[0m'
+
+# Temporary directory
+TEMP_DIR=$(mktemp -d)
+trap 'rm -rf "$TEMP_DIR"' EXIT
+
+# Logging functions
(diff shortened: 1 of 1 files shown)


scripts/check_redirects.sh
```diff
@@ -0,0 +1,184 @@
+#!/bin/bash
+
+# Content API redirect checker - validates redirects and detects daisy chains
```
RubenSandwich, line 3: @im2nguyen can you provide more examples here?

scripts/check_redirects.sh
```diff
@@ -0,0 +1,184 @@
...
+    
+    success "Validation complete: $valid/$total redirects valid"
+}
+
+# Generate reports
+generate_report() {
+    local valid=$(cat "$TEMP_DIR/valid")
+    local total=$(cat "$TEMP_DIR/total")
+    local broken_count=$((total - valid))
+    local daisy_count=$(wc -l < "$TEMP_DIR/daisy" 2>/dev/null || echo "0")
```
RubenSandwich, line 142: Can you be a little more specific on what this variable means?

RubenSandwich (changes requested):
@im2nguyen This is excellent work but I am going to ask for a few changes:

1. Can you rewrite `check_redirects.sh` into a `.mjs` JS file?
  - I'm asking this because:
    - JS is easier to maintain and extend then a shell script
    - Can then testing on top of it
    - Better pass around it's output
  - Cladue Sonnet 4 should be very helpful here
2. Once this is in JS let's make a basic GitHub action that runs the checker if a redirect file is changed.
  - Because right now we can only catch bad redirects in prod, with this we should be able to catch some of them during dev

scripts/check_redirects.sh
RubenSandwich: I'm guessing you mean links that loop back on themselves?

andys-hashi:
Closing this PR. Will revisit at a later time.

PR #832, opened 2025-08-19
Add docs for AES-CBC

rculpepper:


content/vault/v1.21.x (rc)/content/api-docs/secret/transit.mdx
```diff
@@ -890,6 +892,12 @@ will be returned.
   decrypt) could be indicative of a security breach and should not be
   ignored.
 
+- `iv` `(string: "")` - Specifies the **base64-encoded** IV value to use with
+  AES-CBC. The value must be exactly 128 bits (16 bytes) long and the user must ensure that
+  for any given context (and thus, any given encryption key) this IV value is
+  **never reused**. If not provided, a random IV will be generated and prepended
+  to the ciphertext.
```
schavis, line 899:
```suggestion
- `iv` `(string: "")` - Specifies the base64-encoded IV value to use with
  AES-CBC. The value must be exactly 128 bits (16 bytes) long and you must
  ensure that the IV is never reused for any given context or encryption key. If
  you do not provide an IV value, Vault generates and prepends a random IV value
  to the ciphertext.
```
Style correction: write in active voice, avoid "this" as a pronoun, avoid using bold strictly for emphasis

schavis (changes requested):

schavis (approved):

PR #848, opened 2025-08-21
Framework: Add documentation for actions in plugin framework

austinvalle:
This PR introduces documentation for [actions](https://github.com/hashicorp/terraform-plugin-framework/pulls?q=is%3Apr+label%3Aactions+is%3Aclosed) in `terraform-plugin-framework`.

The actual action documentation is just two pages, as we'll likely end up revamping the docs when we add more action types, so didn't want to spread them out too much. Open to suggestions if there is a more natural flow for following the docs.

The syntax of actions is still being determined, but I didn't write any configuration examples here (we can add later if we prefer), so this should be ready to merge when we release `terraform-plugin-framework@v1.16.0`

### Notes
- As there is a new schema, there are updates we have to make to some of the common documentation, although I decided to just make the wording more generic so we don't have to update those pages every time we introduce a new schema. For example, we don't have to list out every single iteration of `BoolAttribute`, so we just list a few 😄 
- I updated the ephemeral resource docs to remove the note about compatibility since it's been released for a while.

### DIFF
--- a/content/terraform-plugin-framework/v1.16.x/docs/plugin/framework/actions/implementation.mdx
+++ b/content/terraform-plugin-framework/v1.16.x/docs/plugin/framework/actions/implementation.mdx
@@ -42,7 +42,7 @@ func (a *DoSomethingAction) Metadata(ctx context.Context, req action.MetadataReq
 
 ## Schema method
 
-The [`action.Action` interface `Schema` method](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/action#Action.Schema) defines a [schema](/terraform/plugin/framework/handling-data/schemas) describing what data is available in the action configuration. An action itself does not have output data that can be referenced by other parts of a Terraform configuration. An action can only plan and apply changes to resource state provided to it, which is declared via the [action type](/terraform/plugin/framework/actions#action-types).
+The [`action.Action` interface `Schema` method](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/action#Action.Schema) defines a [schema](/terraform/plugin/framework/handling-data/schemas) describing what data is available in the action configuration. An action itself does not have output data that can be referenced by other parts of a Terraform configuration.
 
 ### Unlinked action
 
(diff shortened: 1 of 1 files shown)


austinvalle:
Broken link detector is correct, some of these package doc links don't exist yet since we haven't released the Go modules 🙂

content/terraform-plugin-framework/v1.16.x/docs/plugin/framework/actions/implementation.mdx
```diff
@@ -0,0 +1,356 @@
...
+
+Invoke is called by Terraform when an action is ready to be executed, either directly via the CLI or during a `terraform apply` command.
+Terraform calls the `InvokeAction` RPC, in which the framework calls the [`action.Action` interface `Invoke` method](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/action#Action.Invoke). The request contains the configuration supplied to Terraform for the action and the response contains any diagnostics related to the action invocation.
+
+Implement the `Invoke` method by:
+
+1. [Accessing the `Config` data](/terraform/plugin/framework/handling-data/accessing-values) from the [`action.InvokeRequest` type](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/action#InvokeRequest).
+1. Performing logic or external calls for the action.
+1. Sending any [progress messages](#sending-progress-messages) while invoking the action.
+1. Return when the action is finished.
```
stephybun, line 160:
```suggestion
1. [Accessing the `Config` data](/terraform/plugin/framework/handling-data/accessing-values) from the [`action.InvokeRequest` type](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/action#InvokeRequest).
2. Performing logic or external calls for the action.
3. Sending any [progress messages](#sending-progress-messages) while invoking the action.
4. Return when the action is finished.
```

stephybun: Ohhh TIL

PR #850, opened 2025-08-22
Update index.mdx

ram-parameswaran:
Move @include 'kubernetes-supported-versions.mdx' to main k8s page rather than hidden in helm


ram-parameswaran:
Will be raising another PR that encompasses changes for all Vault versions.

PR #854, opened 2025-08-22
Add LDAP auth rotation_url docs

robmonte:


content/vault/v1.21.x (rc)/content/docs/auth/ldap.mdx
```diff
@@ -238,10 +240,17 @@ of the root credential until the field is reset to `false`. If you use
 `rotation_period`, setting `disable_automated_rotation` also resets the credential
 TTL.
 
-For more details on rotating root credentials in the Azure plugin, refer to the
+#### Rotation URL
+
+Use [`rotation_url`](/vault/api-docs/auth/ldap#rotation_url) to provide the specific LDAP
+server URL that the bind account belongs to for rotating its credential. Use this in cases
+such as configuring an LDAP forest where the bind account belongs to only a single domain
+behind a load-balancer. Defaults to the same value as `url`.
```
schavis, line 248:
```suggestion
Use [`rotation_url`](/vault/api-docs/auth/ldap#rotation_url) to provide the
specific LDAP server URL that the bind account belongs to for rotating its
credential. By default, `rotation_url` uses the value of `url`. Setting an
explicit `rotation_url` parameter is useful when you need to configure an LDAP
forest where the bind account belongs to only a single domain behind a
load-balancer. 
```

schavis (changes requested):

PR #874, opened 2025-08-27
Update Azure Secrets Engine API documentation

kevin-loehfelm:
This PR makes the following changes to Azure Secrets Engine HTTP API documentation

- Add documentation for 'Read role'. This has been available but was missing in documentation.
- Update default values and descriptions for 'max_ttl' and 'explicit_max_ttl' in 'Create/update role'. When these values are not explicitly set, they default to 0. The update clarifies that there is a default value and the expected behavior.
- Minor style correction on `ttl`
- JSON formatting on payload sample for 'Create/Update role' on 1.16.x and 1.17.x

This PR updates documentation versions back to the initial LTS release (1.16)

### DIFF
--- a/content/vault/v1.16.x/content/api-docs/secret/azure.mdx
+++ b/content/vault/v1.16.x/content/api-docs/secret/azure.mdx
@@ -252,7 +252,7 @@ $ curl \
 
 The Read Role endpoint returns information about the named role:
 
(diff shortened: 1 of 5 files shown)


kevin-loehfelm:
In reviewing the preview, I noticed version 1.20 was referenced in the description for read role, but it's not version specific. Most recent commit generalizes the statement.

PR #879, opened 2025-08-27
Fix Terrafrom typos

james-warren0:
This is a bulk replace for `Terrafrom` -> `Terraform`.


content/terraform-cdk/v0.14.x/docs/cdktf/api-reference/csharp.mdx
```diff
@@ -11371,7 +11371,7 @@ private object ToTerraform()
 | --------------------------------------------------------------------------------------- | ----------------------------- |
 | <code><a href="#cdktf.TerraformOutput.isConstruct">IsConstruct</a></code>               | Checks if `x` is a construct. |
 | <code><a href="#cdktf.TerraformOutput.isTerraformElement">IsTerraformElement</a></code> | _No description._             |
-| <code><a href="#cdktf.TerraformOutput.isTerrafromOutput">IsTerrafromOutput</a></code>   | _No description._             |
+| <code><a href="#cdktf.TerraformOutput.isTerraformOutput">IsTerraformOutput</a></code>   | _No description._             |
```
ansgarm, line 11374: While we can fix this, it would be technically wrong, since these docs were autogenerated and I just checked and in e.g. v0.14 of CDKTF the typo was actually in the Code:
https://github.com/hashicorp/terraform-cdk/blob/v0.14.0/packages/cdktf/lib/terraform-output.ts#L46

So, yes, obviously wrong, but also these docs are correct – a different place is wrong 😄
(I suppose this came up via the global search and replace)

ansgarm (commented):
We should drop the `api-reference` changes from this PR, as they were generated and the "Terrafrom" in them was actually in the code they describe at that time.

content/terraform-cdk/v0.14.x/docs/cdktf/api-reference/csharp.mdx
james-warren0: Done! I've reverted the changes under `content/terraform-cdk/*/docs/cdktf/api-reference/*`. The `Terrafrom` typo in the `environment-variables.mdx` was still fixed since it was a reference to Terraform Cloud, and not the API function names.

vmacdougal-hc (approved):

xiehan (approved):

hamzahashicorp (approved):
thank you for doing this!

PR #889, opened 2025-09-01
update name from Terraform Cloud to HCP Terraform

andrewwilson5454:
For consistency with other versions and to align with the standard name for Terraform's cloud platform


SarahFrench (approved):
Thanks!

PR #891, opened 2025-09-01
docs(terraform/backend/oss): fix typo on prefix field

clavinjune:


radeksimko (approved):
Good catch, thank you for the PR

PR #910, opened 2025-09-04
Rn/prebuild mdx transforms

RubenSandwich:
- [🎟️ Ticket - run MDX transforms through a pre-build binary](https://app.asana.com/1/90955849329269/project/1210733672911932/task/1211254797572591?focus=true)

## Description

Speeds up prebuild by running a prebuilt prebuild binary compiled with bun.

This did require a few changes:
1. Adding a [run-prebuild.mjs](https://github.com/hashicorp/web-unified-docs/pull/910/files#diff-85df3878a109d83b989186e0b57714feef9f58da7004b6630b21cfa3ca719797) that runs the arch appropriate prebuild binary
1. When creating the [prebuild binary](https://github.com/hashicorp/web-unified-docs/pull/910/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R29), we gzip it or else we go past the 100mb github file size limit
1. Creating a [check-prebuild-binaries.yml GHA file that if any files used by prebuild are changed makes sure the binary is rebuilt](https://github.com/hashicorp/web-unified-docs/pull/910/files#diff-8ec2a0c29001faa144599da90841538ffdd80bc81e4789a0b8f2bd41b3ffab70)

Other things are where not strictly necessary but where done to improve the code/make writing the check-prebuild-binaries.yml GHA easier:
1. Putting most prebuild files in a `/prebuild` folder
1. Updating to use [modern node std lib style imports](https://github.com/hashicorp/web-unified-docs/pull/910/files#diff-136edb01e71b59b4efff6c724b5b0c3af6b0423d78aaa0282a88e26c3c5ee86aL7)
    1. Instead of `import fs from 'fs'`
    1. Using `import fs from 'node:fs'`
1. Using [Node import path resolution](https://github.com/hashicorp/web-unified-docs/pull/910/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R7) so we can use alias's in .mjs files
1. Moved `productConfig.mjs` to the root folder as it's used by both /app and /scripts

## Benchmarks

BEFORE (on main):
```
hyperfine -w 1 -r 3 "npm run prebuild"
  Time (mean ± σ):     139.261 s ±  1.259 s    [User: 127.043 s, System: 31.478 s]
  Range (min … max):   138.336 s … 140.695 s    3 runs
```

AFTER (on this branch):
```
hyperfine -w 1 -r 3 "npm run prebuild"
Benchmark 1: npm run prebuild
  Time (mean ± σ):     69.104 s ±  0.623 s    [User: 58.733 s, System: 12.838 s]
  Range (min … max):   68.478 s … 69.725 s    3 runs
```

Change in execution time: 69.104s / 139.261s = 0.4962
So with these changes `prebuild` is ~0.50% faster and saves ~1 min 10s per run.

## Testing

1. test local dev env
2. test docker dev env
3. poke around on the preview to make sure that the website loads correctly

### DIFF
--- a/scripts/prebuild/run-prebuild.mjs
+++ b/scripts/prebuild/run-prebuild.mjs
@@ -81,10 +81,8 @@ async function runPrebuild() {
 		const args = process.argv.slice(2)
 
 		if (binaryExists) {
-			// For binary files, execute directly
 			execFileSync(filename, args, { stdio: 'inherit' })
 		} else {
-			// For node scripts, execute with node
 			execFileSync('node', [path.join(__dirname, 'prebuild.mjs'), ...args], {
 				stdio: 'inherit',
 			})


tsconfig.json
```diff
@@ -21,9 +22,11 @@
 		],
 		"baseUrl": ".",
 		"paths": {
-			"@utils/*": ["app/utils/*"],
-			"@api/*": ["app/api/*"]
-		}
+			"#utils/*": ["./app/utils/*"],
```
RubenSandwich, line 25: These really should not be needed, but during next.js build them seem to ignore Node import path aliases... 😢

scripts/prebuild/run-prebuild.mjs
```diff
@@ -0,0 +1,98 @@
...
+		filename = `${path.join(__dirname, 'prebuild-x64-linux-binary')}`
+		console.log('Running prebuild binary for x64 Linux...')
+	} else {
+		// Default fallback
+		console.log('Running default node prebuild...')
+	}
+
+	if (binaryExists) {
+		try {
+			if (existsSync(filename)) {
```
RubenSandwich, line 62: Yup we delete the old binary and unzip a new binary each time. It's super duper quick and it means that we **don't** have to worry about folks having old binaries running. Which is a big win.

PR #928, opened 2025-09-09
Fix link to vault license

fleish:
current link (https://developer.hashicorp.com/vault/license) leads to a could not be found page

I believe the intended destination is @ https://developer.hashicorp.com/vault/license which is what this change proposes


schavis:
@fleish Please sign the CLA so we can merge your contribution

PR #957, opened 2025-09-17
Add docs for explorer workspace tags [TF-30537]

KenCox-Hashicorp:
[JIRA](https://hashicorp.atlassian.net/browse/TF-30537)

Explorer workspaces now have a `tags` field, add to documentation.


nylkotuby (commented):
1 opportunity for clarification, but everything here looks right to me and the cURL works on my machine 👍

PR #962, opened 2025-09-18
Adding vault global inclusion exclusions

williamdalessandro:
## Links
[Preview Link](https://unified-docs-frontend-preview-bmja4z3w4-hashicorp.vercel.app/vault/docs) 
[Asana task](https://app.asana.com/1/90955849329269/project/1211223868820586/task/1211058606978186?focus=true)

## What
This is an implementation of version based directives specifically for Vault- differs from the TFC and TFEnterprise exclusions because those were simplified. Does accurate parsing of the version numbers along with relevant error handling/checks. Created new test suite for the new vault exclusions and added in new test to terraform exclusions. Included support for global partials.

## Why
Requested in the above asana ticket

## How
Reworked the entire content exclusion process to be more extensible and to have support for global partials. Added in a README to record the current state of the code and to make it easier to build upon. Reworked the AST parsing/traversal logic to find edge cases/bugs that now have tests to check for them.

## Pics
**The global partial file:**
Screenshot 2025-10-09 at 10 39 25 AM

**The file referenced in 3 different versions (v14, v19, and v20):**
**v14:**
Screenshot 2025-10-09 at 10 39 54 AM
**v19:**
Screenshot 2025-10-09 at 10 40 04 AM
**v20:**
Screenshot 2025-10-09 at 10 40 17 AM

The outputs of each post exclusion:
**v14:**
Screenshot 2025-10-09 at 10 41 32 AM
**v19:**
Screenshot 2025-10-09 at 10 41 18 AM
**v20:**
Screenshot 2025-10-09 at 10 40 50 AM

## Testing
Can run to see the entire test suite

> `npx vitest scripts/prebuild/mdx-transforms/`

To view integration tests with partials+exclusions

> `npx vitest scripts/prebuild/mdx-transforms/build-mdx-transforms.test.mjs`

Or run this to see specific content exclusion tests

> `npx vitest scripts/prebuild/mdx-transforms/exclude-content/index.test.mjs`

## Other info
Had to recompile the binaries since these are changes that take place in the scripts folder


content/vault/v1.20.x/content/docs/concepts/client-count/test-vault-exclusion.mdx
```diff
@@ -0,0 +1,25 @@
+---
+  page_title: Test Vault Content Exclusion
+---
+
+# Test Vault Content Exclusion
+
+This content should always appear.
+
+<!-- BEGIN: VLT:>=v1.21.x -->
```
RubenSandwich, line 9: Let's make sure not to include this file when we ship this PR.

scripts/prebuild/mdx-transforms/build-mdx-transforms.mjs
```diff
@@ -24,9 +24,12 @@ import {
 	loadRedirects,
 } from './rewrite-internal-redirects/rewrite-internal-redirects.mjs'
 import { transformExcludeTerraformContent } from './exclude-terraform-content/index.mjs'
+import { transformExcludeVaultContent } from './exclude-vault-content/index.mjs'
 
 import { PRODUCT_CONFIG } from '#productConfig.mjs'
 
+export const DIRECTIVE_PRODUCTS = ['VLT', 'TFC', 'TFEnterprise']
```
RubenSandwich, line 31: @williamdalessandro let's move this into the exclude-terraform-content transform. That logic is specific to there. I think you moved it here so it's "easier to find" but we should keep all of the logic for the transformed self contained.

scripts/prebuild/mdx-transforms/exclude-terraform-content/index.mjs
```diff
@@ -125,6 +127,12 @@ export function transformExcludeTerraformContent({ filePath }) {
 
 			// TODO: line start and end do not take into account front matter, as it is just tree parsing and technically front matter is not part of the MDX tree
 			if (!directive) {
+				// Check if this is a product we should handle
+				const productMatch = flag.match(/^(\w+):/)
+
+				if (productMatch && DIRECTIVE_PRODUCTS.includes(productMatch[1])) {
```
RubenSandwich, line 133: Hmm I like this, I figure you did it to save build time. It definitely seems to work.

"Deploy Project Artifacts and Set GitHub Outputs" for UDR with this change in CI/CD is 8 mins here versus 11-12 mins normally. 🎉

scripts/prebuild/mdx-transforms/build-mdx-transforms.mjs
RubenSandwich: Looking over this PR a bit more I think I see why you did this. You separated out the TF/TFE/HCP TF exclusions from the Vault exclusions. In that case I think it should be in it's own file; something like so:

```
scripts/prebuild
- ...
- exclude-content
  - constants.mjs
  - exclude-terraform-content.mjs
  - exclude-vault-content.mjs
  - ...
```

scripts/prebuild/mdx-transforms/exclude-vault-content/index.mjs
```diff
@@ -0,0 +1,242 @@
...
+					`Version comparison failed: ${error.message}`,
+					tree,
+				)
+			}
+		})
+		return tree
+	}
+}
+
+const getTfeSemver = (version) => {
```
RubenSandwich, line 208: How come this code is in the vault content exclusion plugin?

RubenSandwich (changes requested):
Let's created a shared.mjs file to keep some of the duplication down and keep the remark transforms contained.

https://github.com/hashicorp/web-unified-docs/pull/962/files#r2361077992

scripts/prebuild/mdx-transforms/exclude-vault-content/index.mjs
williamdalessandro: that function name was an oversight on my part- I used a [similar script](https://github.com/hashicorp/mktg-content-workflows/blob/main/shared/transforms/strip-terraform-enterprise-content.ts#L216) from mktg-content-workflows as a reference. I would like to refactor all this to a cleaner plugin based OOP approach, but for now I went the route of making the code reuse obvious so that the abstraction can be easier to see later on

RubenSandwich: Considering that Transformer is already a subclass, a subclass of a subclass is usually more trouble then it's worth IMO. Let's just make some shared functions.

scripts/prebuild/mdx-transforms/build-mdx-transforms.mjs
RubenSandwich: @williamdalessandro one more thought here. Let's use `Vault` instead just `VLT`. Mainly for the reason that from my understanding we don't never use VLT to mean Vault. Whereas TFC and TFE are both commonly used. It's just a few extra characters that I think will make things a bit clearer for the education team.

scripts/prebuild/mdx-transforms/exclude-vault-content/index.test.mjs
```diff
@@ -0,0 +1,271 @@
...
+				.use(remarkMdx)
+				.use(mockTransform, { filePath, version })
+				.process(markdown)
+			return processor.contents
+		}
+
+		await expect(async () => {
+			return await customRunTransform(markdown, vaultVersion, filePath)
+		}).rejects.toThrow(
+			'Directive block CONSUL:>=v1.15.x could not be parsed between lines 2 and 4',
```
rmainwork, line 268: I wonder if it might be better to do a regex match against the error thrown rather than an exact string match. Just to make the test more robust. That way if the error message changes for casing or punctuation, it won't fail.

scripts/prebuild/mdx-transforms/exclude-content/index.mjs
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Route directive blocks to appropriate processors with if-blocks
+ */
+function routeAndProcessBlock(block, tree, options) {
+	// Parse the directive: "Vault:>=v1.21.x" -> product="Vault", directive=">=v1.21.x"
+	const [product, ...rest] = block.content.split(':')
+	const directive = rest.join(':') // Handle edge cases like "TFEnterprise:only name:something"
+
+	// Explicit routing
+	if (product === 'Vault') {
```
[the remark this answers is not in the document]
williamdalessandro: this change sounds good

scripts/prebuild/mdx-transforms/build-mdx-transforms-file.mjs
```diff
@@ -87,15 +87,28 @@ export async function applyFileMdxTransforms(entry, versionMetadata = {}) {
...
+		const isGlobalPartial = filePath.includes('/global/partials/')
+
+		const processor = remark()
 			.use(remarkMdx)
-			.use(transformExcludeContent, {
+			// Process partials first, then content exclusion
+			// This ensures exclusion directives in global partials are properly evaluated
+			.use(remarkIncludePartialsPlugin, { partialsDir, filePath })
+
+		// Only apply content exclusion if this is NOT a global partial
```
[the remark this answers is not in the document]
RubenSandwich: Everything else in this PR looks great. 🎉 

We just need to double check this before approval.

RubenSandwich (approved):
Looks great @williamdalessandro!

scripts/prebuild/mdx-transforms/exclude-content/ast-utils.mjs
```diff
@@ -100,60 +100,151 @@ export function parseDirectiveBlocks(tree) {
...
 
 		for (let i = 0; i < nodes.length; i++) {
 			const node = nodes[i]
 			const hasPosition = node.position?.start?.line && node.position?.end?.line
 
+			// Recursively check children FIRST to find any nested END comments
+			// before deciding whether to remove this node
+			// const hadChildren = !!node.children
+			const wasInsideRange = insideRange
+			if (node.children) {
```
RubenSandwich, line 122: Thanks for adding tests to cover all of this. It definitely seemed like this rabbit hole went deep. 😢

.gitignore
```diff
@@ -61,3 +61,7 @@ tfe-releases-repos.json
 scripts/prebuild/prebuild-arm-mac-binary
 scripts/prebuild/prebuild-x64-linux-binary
 scripts/prebuild/prebuild-arm-linux-binary
+
+scripts/prebuild/mdx-transforms/benchmarks
+
+scripts/prebuild/mdx-transforms/content-exclusion-OOP-approach
```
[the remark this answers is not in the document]
williamdalessandro: this PR isn't still quite ready- I haven't done the clean up yet so these artifacts will definitely still be littered all throughout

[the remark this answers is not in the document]
RubenSandwich: Oh whoops sorry, it showed up as "ready to review" in my github dashboard.

PR #973, opened 2025-09-22
(terraform query) `-generate-config` -> `-generate-config-out`

koki-develop:
https://developer.hashicorp.com/terraform/cli/v1.14.x/commands/query#generate-import-configuration

```sh
# ❌
$ terraform query -generate-config=...

# ⭕
$ terraform query -generate-config-out=...
```


dbanck (approved):
Nice catch! Thanks for the contribution!

PR #982, opened 2025-09-23
Enhance aws_ami data source configuration (this data source does not …

ooluwgb:
The aws_ami data source does not accept an argument of "id"; it only exports it.


justinretzolk (approved):
Looks good to me 🚀 

[Data Source Example for reference](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami#example-usage)

PR #992, opened 2025-09-24
Document password policy entropy source

sgmiller:
System wide password policies now have a new config parameter,
`entropy_source` which allows customers to pick `seal` if they want entroy
from entropy augmentation.

### DIFF
--- a/content/vault/v1.21.x (rc)/content/api-docs/system/policies-password.mdx
+++ b/content/vault/v1.21.x (rc)/content/api-docs/system/policies-password.mdx
@@ -40,11 +40,13 @@ generation times.
   base64-encoded to avoid string escaping. See [Password Policy Syntax](/vault/docs/concepts/password-policies#password-policy-syntax)
   for details on password policy definitions.
 
-- `entropy_source` `(string: "")` - Specifies the source of entropy
-  (randomness) used to generate the passwords.  The default ("") and
-  "platform" source randomness from the platform RNG.  "seal" (enterprise
-  only) sources entropy from the entropy augmentation.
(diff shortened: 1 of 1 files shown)


content/vault/v1.21.x (rc)/content/api-docs/system/policies-password.mdx
```diff
@@ -40,6 +40,11 @@ generation times.
   base64-encoded to avoid string escaping. See [Password Policy Syntax](/vault/docs/concepts/password-policies#password-policy-syntax)
   for details on password policy definitions.
 
+- `entropy_source` `(string: "")` - Specifies the source of entropy
+  (randomness) used to generate the passwords.  The default ("") and
+  "platform" source randomness from the platform RNG.  "seal" (enterprise
+  only) sources entropy from the entropy augmentation.
```
schavis, line 46:
```suggestion
- `entropy_source` `(string: "platform")` - Specifies the source of entropy
  (randomness) used to generate the passwords.  Must be one of:
    - `platform` - source randomness from the platform RNG.
    - `seal` - source entropy from the entropy augmentation. <EnterpriseAlert inline="true" />
```
It sounds like the default *is* `platform`?

schavis (changes requested):

content/vault/v1.21.x (rc)/content/api-docs/system/policies-password.mdx
sgmiller: The default is unset, because existing policies won't have a value for it.   We could change the readout code to produce platform, but it'd produce Terraform diffs.

schavis (approved):

PR #1013, opened 2025-09-29
correct `-mfa` argument format

Soromeister:
This PR is expected in: [Vault > Docs > Commands: configure-environment-variables **`[-mfa | VAULT_MFA] (string : "")`**](https://developer.hashicorp.com/vault/docs/commands#configure-environment-variables)

Where the present example (in the image below) as well as the details have been updated & improved.
image

Should be **back-ported** for older versions of the docs too.


mister-ken (approved):
lgtm

PR #1015, opened 2025-09-30
Docs: Update unclear statement in how-terraform-works.mdx

Andersson007:
Hello. I'm reading the documentation the first time and the statement `Terraform Plugins expose IMPLEMENTATION` doesn't feel right to me in this context. IMO, plugins typically not expose, but provide or encapsulate implementation for interacting with a specific service.

Here's alternative wording.


Andersson007:
the second CLA signing doesn't seem to work.. The first is green though

PR #1021, opened 2025-09-30
content/terraform: fix missing codeblock closure

jar-b:
Adds a missing codeblock closure causing the `module` configuration reference to render partially invalid HCL.

image


rkoron007 (approved):
Thanks for catching!

PR #1036, opened 2025-10-02
Correct action_trigger event syntax (remove quotes)

nphilbrook:
The example action_trigger here is incorrect, the event types *cannot* be quoted.


BrianMMcClain (approved):
Good catch, thanks!

PR #1045, opened 2025-10-02
Compress TFE images

RubenSandwich:
The initial commit is limited to TFE v1.0.x, as lots of images are replicated across all TFE versions. Leading to ~7.5k images total. So by limiting to just TFE v1.0.x we only have 136 to compare.

Care has been taken to make sure that no noticeable differences have been introduced through compression.

Once the visual compare has passed education the rest of the TFE images will be compressed as well.


nikolasrieble (approved):
Thank you very much!

PR #1048, opened 2025-10-03
Update labeler for hcp-docs support

jonathanfrappier:
Update labeler.yml to support hcp-docs migration
- Adds remaining HashiCorp products to support both `hcp` and `<productname>` label
- Adds assumed path for non HCP product docs (which can be edited when those products migrate)


RubenSandwich (approved):
🚢

PR #1053, opened 2025-10-05
Remove K8s Auth Requirement From Docs

jaireddjawed:
We no longer plan to require audiences in the kubernetes auth plugin. It was later discovered that doing so could break some valid configurations of our customers (more details [here](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/300#issuecomment-3179814054)). 

Instead, we plan to simply recommend that customers specify an`audience` if it does not break their workflow. We log a warning in Vault when an audience is not configured so that customers will be aware.

### DIFF
--- a/content/vault/v1.20.x/content/docs/updates/important-changes.mdx
+++ b/content/vault/v1.20.x/content/docs/updates/important-changes.mdx
@@ -210,8 +210,8 @@ $ vault write auth/kubernetes/role/demo      \
 
 You would then authenticate with the command below.
 
-```
-vault write auth/kubernetes/login role=demo audience="my_audience" jwt=...
+```shell-session
+$ vault write auth/kubernetes/login role=demo audience="my_audience" jwt=...
 ```
 
 ---


schavis:
@jaireddjawed If we're updating the current docs, the PR should merge to main instead of the release branch :)

content/vault/v1.20.x/content/docs/updates/important-changes.mdx
```diff
@@ -192,19 +192,27 @@ audience.
...
-We recommend updating your policies before Vault makes the audience value
-required for all Kubernetes authentication roles so Vault can explicitly
-validate that the audience claim in JWT tokens (`aud`) is intended for Vault and
-not another service. For example:
+We recommend configuring an audience value for Kubernetes authentication roles when possible.
+This enables Vault to validate that the aud (audience) claim in JWT tokens is explicitly intended for Vault,
+reducing the risk of token misuse by other services. While not required, setting an audience is considered a security
+best practice and is encouraged when it doesn't interfere with your workflow. For example, tokens created using
+kubernetes.io/service-account-token do not include an aud claim, so it's appropriate to omit the audience field in this case.
+Here's an example configuration.
```
schavis, line 200:
```suggestion
There are cases where configuring audience details can interfere with your
workflow. For example, tokens created using `kubernetes.io/service-account-token`
do not include an aud claim. But we recommend configuring an audience value for
Kubernetes authentication roles whenever possible. Setting explicit audience
details is best practice because it reduces the risk of token misuse by other
services. Vault can use the configured values to validate that the `aud`
(audience) claim in JWT tokens is intended for Vault.

For example:
```
Style correction: avoid "this" as a pronoun, avoid possessives

schavis (changes requested):

PR #1058, opened 2025-10-06
Remove K8s Auth Requirement From Docs

jaireddjawed:
We no longer plan to require audiences in the kubernetes auth plugin. It was later discovered that doing so could break some valid configurations of our customers (more details [here](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/300#issuecomment-3179814054)). 

Instead, we plan to simply recommend that customers specify an`audience` if it does not break their workflow. We log a warning in Vault when an audience is not configured so that customers will be aware.


schavis (commented):
For future reference, you can just change the base (target) of the PR, we don't require folks to create entirely new PRs :)

PR #1068, opened 2025-10-07
Add a note to change the `tfc_hostname` variable for EU orgs

rkoron007:
Slack thread: https://hashicorp.slack.com/archives/C028C7W7G4T/p1758832361751299

### DIFF
--- a/content/vault/v1.19.x/content/api-docs/secret/databases/index.mdx
+++ b/content/vault/v1.19.x/content/api-docs/secret/databases/index.mdx
@@ -551,7 +551,7 @@ this in order to know the password.
 - `self_managed_password` `(string)` – <EnterpriseAlert product="vault" inline />
   The password corresponding to the username in the database. Required when using
   the Rootless Password Rotation workflow for static roles. Only enabled for select
-  DB engines (Postgres). This parameter is deprecated in favor of `password`.
+  DB engines (Postgres, Oracle). The `self_managed_password` parameter is deprecated in favor of `password`.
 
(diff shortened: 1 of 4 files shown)


content/terraform-docs-common/docs/cloud-docs/integrations/aws-service-catalog/index.mdx
```diff
@@ -27,6 +27,16 @@ By combining HCP Terraform with AWS Service Catalog, we’re combining a self-se
 
 ## Installation & Configuration
 
+<!-- BEGIN: TFC:only name:hcp-eu -->
+
+<Note>
+
+To use the AWS service catalog integration with an HCP Europe organization, set the example [`tfc_hostname` variable](https://github.com/hashicorp/aws-service-catalog-engine-for-tfc/blob/main/variables.tf#L15-L19) to `app.eu.terraform.io` in your configuration.
+
+</Note>
```
trujillo-adam, line 36:
```suggestion
To use the AWS service catalog integration with an HCP Europe organization, set the example [`tfc_hostname` variable](https://github.com/hashicorp/aws-service-catalog-engine-for-tfc/blob/main/variables.tf#L15-L19) to `app.eu.terraform.io` in your configuration.
```
What's the reason for calling this out as a note?

trujillo-adam (approved):
I suggested taking this info out of the note alert, but I defer to you. 
I think we should also replace the ampersands in the section  heading with "and". That is, if we have to combine "Installation" and "Configuration" into a single step.

PR #1102, opened 2025-10-13
Stacks - specify to delete the component block while adding `remove`

rkoron007:
Ticket: https://hashicorp.atlassian.net/browse/IPE-1315


nfagerlund (approved):
Awesome, I like that reframing a lot.

PR #1120, opened 2025-10-15
VAULT-39667 Add docs for breaking changes to allowed_ and denied_parameters

bosouza:
This PR adds documentation about Vault's breaking change to the `allowed_parameters` and `denied_parameters` behavior when comparing list parameters.

Jira: [VAULT-39667](https://hashicorp.atlassian.net/browse/VAULT-39667)
Internal draft: [Documentation for denied_parameters breaking change](https://docs.google.com/document/d/1FJldPo9Nc2wNjk3j5Zp-sTIJu1yxMCX1ergHrijLd70/edit?tab=t.0)

[VAULT-39667]: https://hashicorp.atlassian.net/browse/VAULT-39667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


content/vault/global/partials/policies/denied_parameters_warning.mdx
```diff
@@ -0,0 +1,5 @@
+!> **Warning:** In Vault versions 1.20 and earlier, attributes of type list had a different behavior
```
schavis: Suggest removing this partial

content/vault/v1.21.x (rc)/content/docs/concepts/policies.mdx
```diff
@@ -485,6 +492,12 @@ constrain requests, using the following options:
   - If any parameters are specified, all non-specified parameters are allowed,
     unless `allowed_parameters` is also set, in which case normal rules apply.
 
+  - Request parameters that are defined in `denied_parameters` and set to a list
+  will be denied if any values in the list are denied, or if the list as a whole
+  is denied. On Vault 1.20 and below only the latter is true, and during the
+  deprecation process Vault can be configured to use the old behavior by setting
+  the environment flag `VAULT_LEGACY_EXACT_MATCHING_ON_LIST` on the server.
```
schavis:
```suggestion
  - Vault allows request parameters defined as lists in `allowed_parameters`
    if the policy allows all the values provided in the list or the list as a
    whole. Vault 1.21.x deprecated exact list-only matching. To force legacy
    behavior during the deprecation process, set the environment flag
    `VAULT_LEGACY_EXACT_MATCHING_ON_LIST` on your Vault server.
```

content/vault/v1.21.x (rc)/content/docs/updates/important-changes.mdx
```diff
@@ -49,6 +49,38 @@ vault write auth/kubernetes/role/demo \
...
+Prior to Vault 1.21, when a parameter in a request was subject to
+`allowed_parameters` or `denied_parameters` checks and the parameter's value was
+a list, Vault evaluated the **entire list** as a single value.
+This meant Vault would only find a match if the **whole list** exactly matched
+an entry in `allowed_parameters` or `denied_parameters`, which was often
+unintuitive.
+
+Starting with Vault 1.21, this behavior has changed.
+When a parameter's value is a list, **each element** in that list is now checked
+individually against the allowed or denied values.
```
schavis:
```suggestion
Vault now checks each value in a list parameter against allowed/deined values
in the applicable Vault policy and allows or denies requests if the policy
defines the list as a whole or every individual element of the list. For
example, if the request includes a list like `['a', 'b', 'c']`, it matches to a
policy that includes `['a', 'b', 'c']` or the individual values `a`, `b`, and
`c`.

Previous versions of Vault only matched list parameters when the associated
policy defined the list as a whole. As a result, Vault allowed lists containing
denied values as long as the policy did not deny that exact list and denied
lists containing allowed values because the policy did not allow the exact list.
```

content/vault/global/partials/policies/allowed_parameters_warning.mdx
```diff
@@ -0,0 +1,4 @@
+!> **Warning:** In Vault versions 1.20 and earlier, attributes of type list had a different behavior
+with `allowed_parameters` and `denied_parameters`. In order for the list to be considered
+allowed, it didn't matter that all values in the list were individually allowed, as the
+whole list had to be defined as a single allowed value. This has been changed in Vault 1.21.
```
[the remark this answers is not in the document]
bosouza: I don't think this table structure helps with explaining the change... I can certainly make a single paragraph to be put under this [parameter constraints ](https://developer.hashicorp.com/vault/docs/concepts/policies#parameter-constraints) section tho:

```mdx
<Warning title="Parameter list behavior">

In Vault versions 1.20 and earlier, parameters of type list were evaluated as a single value when checked against `allowed_parameters` and `denied_parameters`.

* For `allowed_parameters`, a list was only considered allowed if the entire list exactly matched an allowed value, even if all its individual elements were allowed.

* For `denied_parameters`, a list was only considered denied if the entire list matched a denied value, so lists containing denied elements could still pass.

Starting in Vault 1.21, each element in a list is now evaluated individually, so lists containing denied values are correctly rejected, and lists containing only allowed values are accepted as expected.

</Warning>
```

content/vault/v1.21.x (rc)/content/docs/concepts/policies.mdx
```diff
@@ -432,6 +432,13 @@ constrain requests, using the following options:
       }
     }
     ```
+  
+  - Request parameters that are defined in `allowed_parameters` and set to a list 
+  will be allowed if all the values in the list are allowed, or if the list as a
+  whole is allowed. On Vault 1.20 and below only the latter is true, and during
+  the deprecation process Vault can be configured to use the old behavior by setting
+  the environment flag `VAULT_LEGACY_EXACT_MATCHING_ON_LIST` on the server.
```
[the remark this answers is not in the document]
bosouza: I rewrote it in active voice (hopefully :D) but without starting with "Vault allows", as I feel like that makes the sentence much harder to understand, please take a look and lmk if you like it better

content/vault/v1.21.x (rc)/content/docs/updates/important-changes.mdx
bosouza: >and allows or denies requests if the policy
defines the list as a whole or every individual element of the list

that's not quite right, vault will allow if all values in the list match the allowed values, but will deny if **any** values in the list match a denied value. That same caveat also makes it hard to smoothly use the example, not sure how to integrate it nicely into the paragraph...

bosouza: sorry, this version doesn't mention the fields to which the breaking change applies at all, I think that's very important.

also
>allows or denies requests if the policy defines the list as a whole or every individual element of the list

the "or every element of the list" is just for `allowed_parameters`, for `denied_parameters` it's "or any element of the list"

content/vault/global/partials/policies/list-allowed-parameters.mdx
```diff
@@ -0,0 +1,24 @@
+#### List parameter evaluation
+
+In Vault v1.20 and earlier, when a request parameter carrying a list value matches
+an `allowed_parameters` or `denied_parameters` policy rule, Vault checks for an exact
+match between each allowed or denied value in the policy and the **entire** list value in
+the request parameter. This behavior can produce unexpected results.
```
schavis:
```suggestion
When a request parameter carrying a list value matches an
`allowed_parameters` or `denied_parameters` policy rule, Vault checks for
an exact match between each allowed or denied value in the policy and
the **entire** list value in the request parameter. The fact that Vault
looks for matches to the entire list can produce unexpected results.
```

Since we only include this partial for docs <= 1.21, we don't need to call out the specific versions

content/vault/global/partials/policies/list-allowed-parameters.mdx
```diff
@@ -0,0 +1,24 @@
...
+one with parameter `"X": ["A", "B"]`, `"X": ["B", "A"]` or even `"X": ["A"]`, because
+none of those list values exactly match one of the allowed values "A" or "B".
+
+The same logic applies to `denied_parameters`: if it is set to `"Y": ["C", "D"]`,
+Vault **allows** a request with parameter `"Y": ["C", "D"]`, which can lead to
+unauthorized access if the intent was to deny any request that includes "C", "D"
+or both. As a result, we **strongly recommend** using `allowed_parameters` instead of
+`denied_parameters` for list parameters.
+
+Additionally, Vault v1.20 and earlier does not treat comma-separated strings in request
```
schavis:
```suggestion
Additionally, Vault does not treat comma-separated strings in request
```

schavis (approved):

PR #1139, opened 2025-10-19
docs: missing quote

mbarczyk:
Missing closing quotation:
```
resource "aws_instance" "example {    <- HERE
  count = 2
  #. . .
}
```


radeksimko (commented):
Great catch, thank you for the PR.

Would you mind fixing the same issue in 1.12.x?

https://github.com/hashicorp/web-unified-docs/blob/2b03006840861f7501a4d9a19776984235c2a1e0/content/terraform/v1.12.x/docs/language/import/index.mdx?plain=1#L87

PR #1166, opened 2025-10-22
Update CODEOWNERS for HCP Vault

jonathanfrappier:
Update CODEOWNERS to make vault-education-approvers default for HCP Vault* docs and partials


RubenSandwich (approved):
🚢

PR #1193, opened 2025-10-28
updated the self-signed certificate generation example

StamatisChr:


connorfryar:
Added a couple of single line comment. 

Context:
Providing the OpenSSL commands like we are already seems like a massive extension of an olive branch for our customers. 

I think it would be appropriate to potentially change the names of the cert and the key to match the start up name requirements of TFE. cert.pem/key.pem. 

Check single line comments for further clarity.

PR #1198, opened 2025-10-29
Add TFE exclusion tags to module test generation

BrianMMcClain:
Module test generation is not supported in TFE. This PR adds the appropriate exclusion tags and backports it to all affected versions of the TFE docs.

### DIFF
--- a/content/terraform-docs-common/redirects.jsonc
+++ b/content/terraform-docs-common/redirects.jsonc
@@ -543,6 +543,28 @@
     "destination": "/terraform/language/v:version/validate",
     "permanent": true
   },
+  // Custom conditions → validate
+  {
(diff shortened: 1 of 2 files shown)


rkoron007 (approved):
Looks good, but see my other comment for the missed versions that we should add this to!

BrianMMcClain:
> I'm so sorry to be that guy - but I think this goes _even_ further back a bit. Looks like it was first introduced in v202312-1 ([preview link](https://unified-docs-frontend-preview-h9vn2c7g5-hashicorp.vercel.app/terraform/enterprise/v202312-1/registry/publish-modules#sharing-modules-across-organizations))

Oh good catch! I'll get these updates as well, my search apparently missed these :(

PR #1202, opened 2025-10-29
Add 1.21.x GUI kvv2 known issues

hellobontempo:
Adds known issues for 1.21 GUI kvv2 metadata list failures

### DIFF
--- a/content/vault/v1.21.x/content/docs/updates/important-changes.mdx
+++ b/content/vault/v1.21.x/content/docs/updates/important-changes.mdx
@@ -92,10 +92,8 @@ filters you have two options:
 
 #### Issue
 
-Users cannot list KV v2 secrets in the GUI if their policy grants `list` access to metadata 
-but the policy path does not include a trailing slash.
-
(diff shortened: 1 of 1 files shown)


content/vault/v1.21.x/content/docs/updates/important-changes.mdx
```diff
@@ -81,3 +81,59 @@ filters you have two options:
...
+
+```
+path "secret/metadata/:path" {   
+  capabilities = ["list"] 
+}
+```
+
+#### Workaround
+
+To resolve the policy issue, add a trailing slash to the policy path until the fix is released:
```
schavis:
```suggestion
Option 1: Add a trailing slash to the policy path:
```
We try to avoid hand-waving at future behavior/fixes.

hellobontempo: yeah, i'm not sure this is a recommended solution as it is, so I think I'll swap it to `option 2` as well

PR #1211, opened 2025-10-30
Update sentinel override docs

andrewassaf7:
### PR Description

This small doc update adds a note to clarify why some mandatory Sentinel policies can still be overridden in Terraform Enterprise.

We had a few customer tickets where users were confused about why “hard mandatory” policies could still be overridden. The behaviour is expected, when the policy set has “This policy set can be overridden in the event of mandatory failures” enabled, users with the right permissions (like admins or team owners) can override both soft- and hard-mandatory policy failures.

This doc change makes that clearer to help prevent future confusion.

**What’s Changed**

- Added a short clarification about when and why overrides can still occur.

**Why**

- To address customer feedback and help users better understand how enforcement levels actually behave in practice.

### DIFF
--- a/content/terraform-docs-common/docs/cloud-docs/workspaces/policy-enforcement/manage-policy-sets/index.mdx
+++ b/content/terraform-docs-common/docs/cloud-docs/workspaces/policy-enforcement/manage-policy-sets/index.mdx
@@ -60,11 +60,15 @@ You can set an enforcement level for each policy that determines what happens wh
 
 ### Sentinel
 
-Sentinel provides three policy enforcement levels:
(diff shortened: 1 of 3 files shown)


andrewassaf7:
> @andrewassaf7 Thanks for adding this information. You added this to the 202301-1 version of the page. Is this the only version that this information applies to? Does this also apply to HCP Terraform?

@trujillo-adam Thanks for catching that! The information applies to all current versions of Terraform Enterprise where Sentinel policy sets support the “This policy set can be overridden in the event of mandatory failures” option. The behaviour is also consistent in HCP Terraform, since the enforcement and override mechanisms are shared across both platforms.

trujillo-adam (approved):

PR #1224, opened 2025-10-31
add redir for import block older than 1.11

trujillo-adam:
This PR adds a redirect for  pre-import block URLs. To verify, go to /terrform/language/block/import, then choose 1.11 or older from the version switcher.


rkoron007 (approved):
👻

PR #1246, opened 2025-11-05
Update Azure Static Secrets docs

Zlaticanin:


content/vault/v1.21.x/content/api-docs/secret/azure.mdx
schavis:
```suggestion
   Vault merges the path metadata with metadata stored with the role, but
   role metadata always takes precedence when keys overlap. Key/value pairs
   appear as strings separated by the `=` symbol (e.g. `key=value`). When
   using the Vault CLI, you can provide multiple tags in the role
   configuration by adding multiple `metadata` assignments in the same command.
```

schavis:
```suggestion
- `skip_import_rotation` (`bool`, `false`) - Determines whether Vault 
   generate credentials automatically when creating a role. Set `skip_import_rotation`
   to `true` if you do not want Vault to create initial credentials on role
   creation. If you skip credential creation, you can generate them later manually
   rotating to role with the `rotate-role` parameter or by importing the
   credentials from a trusted source.
   If you omit `skip_import_rotation` or set it to `false`, Vault automatically
   provisions the first set of credentials during role creation.
```

Zlaticanin:
closing in favor of: https://github.com/hashicorp/web-unified-docs/pull/1285

PR #1275, opened 2025-11-11
terraform 1.14 search ga

trujillo-adam:
**Do not merge until Nov 19.**
 
This PR removes beta language and text in Terraform language and HCP Terraform docs for the GA of search and import functionality per https://hashicorp.atlassian.net/browse/IPE-1288.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -67,3 +67,5 @@
 #Vagrant
 /content/vagrant @hashicorp/Vagrant
 
+# Nomad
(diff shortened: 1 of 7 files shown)


BrianMMcClain (approved):
This LGTM! Only thing I noticed is that we had previously added aliases in the nav data files in the 1.13 docs to help users find the import, query, etc. pages when using language used to market the feature (such as "search") which I don't see in the 1.14 docs. Do we want to add those to this version as well?

PR #1289, opened 2025-11-12
Add Docker network and extra hosts configuration

straubt1:
These configurations worked for standing up TFE, however when a Run was created and an agent container spawned, it was not created in the same Docker Network as Terraform Enterprise.

These changes were confirmed to work for Disk and External mode (I did not test for Active-Active).


krastin:
Hi @straubt1 !

I see your issue, and I reckon it's due to the fact that the tutorials assumes you are deploying in the default Docker network.
It would be equally inaccurate to assume that the deployment needs to go into the default Terraform Enterprise Docker network.

What I believe needs to happen here is to point out in the tutorial that the deployment is targeted at the default Docker network, and if anyone needs to alter this, you can do that with the help of the `TFE_RUN_PIPELINE_DOCKER_NETWORK` environment variable. I will close this PR and implement this change in [the following PR](https://github.com/hashicorp/web-unified-docs/pull/3534). The description about `TFE_RUN_PIPELINE_DOCKER_EXTRA_HOSTS` is already discussed on the same page, even if not in the example deployment YAML code.

Nonetheless - thank you for the PR submission and for taking the time to raise this issue!
Krastin

PR #1358, opened 2025-11-20
Compress vagrant images

rmainwork:
Compress Vagrant images.

No great rush for this, but it is an item on the migration checklist...

---
  - https://app.asana.com/0/0/1211925524859020


RubenSandwich (approved):
👍🏻

PR #1367, opened 2025-11-21
Simplify dev-portal redeployment

RubenSandwich:
[Ticket - UDR deployments fail to consistently kick off necessary dev-portal builds](https://app.asana.com/1/90955849329269/project/1205890531436279/task/1211735519867470?focus=true)

## Description

This PR simplifies the deployment of dev-portal when UDR changes. It does so by fully reloading dev-portal on every change. This is a bit of a heavy hand approach, [with a more fine-grain approach being working on here](https://github.com/hashicorp/web-unified-docs/pull/1250), but we are taking this approach to get this work out ASAP.

## Testing

Because of the small and simplified nature of this change/code, I do not believe testing is necessary. Just a manual review from another engineer.


williamdalessandro (approved):
Cool! I didn't know about those deploy hooks

PR #1372, opened 2025-11-21
Fix GA-RC assumptions about release branch

schavis:
- Change the `-branch` flag to capture explicit release branch 
- Add a GA brach flag to capture explicit GA branch
- Change rcBranch logic to use the new flag
- Fix a string error in the bash helper that was injecting double quotes into the file path


RubenSandwich (approved):
👍🏻

PR #1378, opened 2025-11-24
Terraform 1.10+: Clarify `providers mirror -lock-file` defaults and value meanings

radeksimko:
While the existing wording isn't strictly wrong, it could be confusing on first read. This PR proposes an update that should make it clearer what the default is and what the values (true or false) mean when set.


mildwonkey (approved):
👍🏻 it's not any clearer to me, but it's not less clear, so LGTM!!!

PR #1383, opened 2025-11-24
Generate unique algolia object IDs for versioned and non versioned products

rmainwork:
This PR fixes a bug reported in our slack support channel where some pages were completely missing from Algolia(most notably [this one](https://developer.hashicorp.com/vault/docs/auth/spiffe) and [this one](https://developer.hashicorp.com/boundary/docs/configuration/kms/ibmkp). This is because the file locations were being split along the directory separator and then array element `1` was being used as the object ID. For non versioned products like WAF, this worked fine - however for products with an extra directory in the path (i.e: vault) this caused the objectID to evaluate to `"docs_vault/vault/v1.20.x"`. Since this is not a unique value, records were overwritten. 

This PR fixes the objectID generation to use `.at(-1)` [here](https://github.com/hashicorp/web-unified-docs/pull/1383/files#diff-c8c265e4c361c96e131e23170b155c96d3ab2a6510496c561bf72856f5bdeffdR49) to always retrieve the last element...which _should_ always be correct for both cases.

Since this PR is somewhat tricky to test in the preview, I've included screenshots below showing the missing records in the staging algolia index. 

image

image

---
  - https://app.asana.com/0/0/1211800252499419


williamdalessandro (approved):
This is a sweet fix @rmainwork ! I'm looking at the diff between the built algolia indexes before and after your changes and noticed it fixes the indexes for like 3.4k+ objectids from boundary, consul, nomad, hcp-docs, etc

Screenshot 2025-11-24 at 6 22 12 PM

Screenshot 2025-11-24 at 6 19 59 PM

PR #1477, opened 2025-12-05
Add images and example for versioned redirects

rkoron007:
Adding images and an example for the redirects file based on my 404 ramblings! I'll eventually make more updates when I'm down with the script, but wanted to add these for now! 

[You can see the rendered file here 🔍 ](https://github.com/hashicorp/web-unified-docs/blob/989bc12ef602c7a99d88762a765e452e12914d8d/docs/content-guide/redirects.md)

### DIFF
--- a/docs/content-guide/redirects.md
+++ b/docs/content-guide/redirects.md
@@ -38,7 +38,6 @@ The three redirects cover the following use cases:
 
 Refer to [Versioned redirects](#versioned-redirects) for a full example.
 
-
 ## Definitions
 
 - **Docset** - A collection of docs associated with a specific product and


docs/content-guide/redirects.md
```diff
@@ -17,6 +17,29 @@ Use redirects for fun and profit!
...
+
+This redirect ensures any requests to the old URL properly re-route to the new URL. To learn more, refer to [Example redirects](#example-redirects).
+
+If your product **is** versioned, add **three** redirects to reroute requests to the appropriate path in each version:
+
+![Diagram showing three redirects for a versioned product move](../images/three_redirects_per_versioned_move.png)
+
+The three redirects cover the following use cases:
+1. Redirects the latest version URL, which has no version path (i.e., `/product/<path>`).
+2. Redirects older versions to the old path (i.e., `/product/:version/<old_path>`).
```
schavis: It's probably worth being explicit that your mapping the *new* path to the *old* path and noting that these redirects are needed explicitly for the version selector. Otherwise it kind of reads like you're creating redirects for old paths that haven't changed.

docs/content-guide/redirects.md
```diff
@@ -17,6 +17,29 @@ Use redirects for fun and profit!
...
+3. Redirects current and future versions to the new path (i.e., `/product/:version/<new_path>`).
+
+Refer to [Versioned redirects](#versioned-redirects) for a full example.
+
+We also recommend that you replace references to the old URL with the new URL in your documentation text.
```
schavis: This is true regardless of whether your content is versioned. I'd suggest adding this to the very beginning in the vein of: "In addition to changing in-content references when a UR changes blah blah blah"

docs/content-guide/redirects.md
```diff
@@ -263,6 +285,41 @@ Redirect any `/old/path/` path that starts with "hello" and ends with "goodbye":
   }
 ```
 
+### Versioned redirects
+
+In the following example, you want to update the URL `/terraform/state` to `/terraform/concepts/state` in Terraform v1.10.x and above. Meaning, you want to move `state.mdx` to a new `concepts` directory.
+
+![Diagram showing example of versioned redirects](../images/versioned_redirects_example.png)
+
+To properly cover this URL change, you need three redirects:
```
schavis:
```suggestion
To properly cover the change to `terraform/concepts/state`, you need three redirects:
```

docs/content-guide/redirects.md
```diff
@@ -263,6 +285,41 @@ Redirect any `/old/path/` path that starts with "hello" and ends with "goodbye":
...
+1. The first redirect is for the latest version, you want `/terraform/state` to redirect to the new `/terraform/state/concepts` URL.
+1. The second redirect is backward-facing. In v1.9.x and below, you want users to land on the original URL `/terraform/<v1.9.x_and_below>/state`.
+1. The third redirect is forward-facing. In v1.10.x and above, you want users to land on the new URL `/terraform/<v1.10.x_and_above>/concepts/state`.
```
schavis:
```suggestion
1. A forward-facing redirect to map URLs for v1.10.x and above so users land on
   the new URL `/terraform/<v1.10.x_and_above>/concepts/state`.
```
Style suggestion: same as (1)

docs/content-guide/redirects.md
```diff
@@ -17,6 +17,29 @@ Use redirects for fun and profit!
...
+
+The single redirect ensures any requests to the old URL properly re-route to the new URL. To learn more, refer to [Example redirects](#example-redirects).
+
+If your product **is** versioned, add **three** redirects to reroute requests to the appropriate path in each version:
+
+![Diagram showing three redirects for a versioned product move](../images/three_redirects_per_versioned_move.png)
+
+The three redirects cover the following use cases:
+1. Redirects the latest version URL, which has no version path (i.e., `/product/<path>`).
+2. Redirects older versions to the old path (i.e., `/product/:version/<old_path>`).
```
rkoron007:
```suggestion
1. Redirects any requests of the new path to the previous path in older versions (i.e., `/product/:old_version/<old_path>`).
```

docs/content-guide/redirects.md
```diff
@@ -17,6 +17,29 @@ Use redirects for fun and profit!
...
+3. Redirects current and future versions to the new path (i.e., `/product/:version/<new_path>`).
```
rkoron007:
```suggestion
1. Redirects any requests from the old path to the new URL in versions where that new path exists (i.e., `/product/:new_version/<old_path>`).
```

docs/content-guide/redirects.md
```diff
@@ -17,6 +17,29 @@ Use redirects for fun and profit!
 
 ---
 
+## Background
+
+If you move an existing article in the documentation, you **must** add redirects. Otherwise, links break between versions and users cannot find the information they are looking for, even if it exists in the version they are viewing.
```
rkoron007:
```suggestion
If you move an existing article in the documentation, you **must** add redirects. Otherwise, links break between versions and users cannot find the information they are looking for, even if it exists in the version they are viewing. Additionally, we also recommend replacing references to the previous URL with your new URL in your documentation text.
```

schavis (approved):

PR #1546, opened 2025-12-22
HCP Vault Dedicated - updating AWS and Azure feature parity.

anurag925:
Description: Updates the HCP Vault Dedicated documentation to reflect that Azure now supports 30-day snapshot retention after cluster deletion.
Changes Made: Removed snapshot retention from the Azure exceptions list since this feature is now supported on Azure.
Impact: Keeps the docs accurate so users aren't misled that snapshot retention is unsupported on Azure when it is actually available and supported.
Screenshot 2025-12-22 at 12 10 59 PM


harinimhc (approved):
LGTM, thanks for raising this @anurag925

PR #1587, opened 2026-01-08
RADAR:-870: Vault Radar Public API for Events Docs

trentdibacco:
Adding documentation for the public api for events.

* Created new section: Public APIs
* Created new page under that section: events.mx

I still need to publish the swagger and enable the feature flag in LD, but I would like to get the review of the doc in place before doing those things.

### DIFF
--- a/content/hcp-docs/content/partials/hcp-api/token-prereq.mdx
+++ b/content/hcp-docs/content/partials/hcp-api/token-prereq.mdx
@@ -1,1 +1,1 @@
-- Retrieve an [access token](hcp/docs/hcp/api#authenticate-to-hcp) for authentication.
+- Retrieve an [access token](/hcp/docs/hcp/api#authenticate-to-hcp) for authentication.


content/hcp-docs/content/docs/vault-radar/api/events.mdx
```diff
@@ -0,0 +1,331 @@
...
+- **Filtering**: Apply conditions using Event field names
+- **Sorting**: Order results by Event field names
+- **Grouping**: Aggregate events by a specific Event field names
+
+### Search Parameters
+
+- **limit** (int32): Maximum number of results to return (max: 1000)
+- **page** (int32): Page number for pagination (first page is 1)
+- **filters** (array of FilterV2): Filtering criteria
+- **orderBy** (string): Comma-delimited Event field name to sort by (e.g., `"created"`, `"severity"`, or multiple order
```
kathytong-hashicorp: nit: Do we want to specify optional for these 3?

content/hcp-docs/content/docs/vault-radar/api/events.mdx
```diff
@@ -0,0 +1,331 @@
...
+
+# Search with groupBy for event counts by severity and status.
+curl -X POST \
+  https://api.cloud.hashicorp.com/2023-05-01/vault-radar/projects/PROJECT_ID/events/search \
+  -H "Authorization: Bearer $TOKEN" \
+  -H "Content-Type: application/json" \
+  -d '{
+    "search": {
+      "filters": [
+        {"id": "in_latest_version", "value": [true], "op": "EQ"}
```
kathytong-hashicorp: Also out of curiosity, for a filter on a field that's type string, if we pass in both null and empty string in `filter.value`, on the Go side will it be able to pass both along into the Python endpoint?

trentdibacco: This works.

content/hcp-docs/content/docs/vault-radar/api/events.mdx
```diff
@@ -190,13 +190,13 @@ The Search Events endpoint uses `SearchSchemaV2` to define search criteria. This
...
 - **Sorting**: Order results by Event field names
-- **Grouping**: Aggregate events by a specific Event field names
+- **Grouping**: Aggregate events by specific Event field names
 
 ### Search Parameters
 
-- **limit** (int32): Maximum number of results to return (max: 1000)
-- **page** (int32): Page number for pagination (first page is 1)
-- **filters** (array of FilterV2): Filtering criteria
+- **limit** (int32): Maximum number of results to return (max: 1000). Optional, defaults to 1000.
```
kathytong-hashicorp: smol nit: I think everything might actually be optional here 😅  For some reason when I made the comment last round I was thinking about `groupBy`, `orderBy`, and `orderDirection` but yeah technically everything is optional sorry about that 😅

kathytong-hashicorp (approved):

PR #1720, opened 2026-01-28
fix typo

maxluchterhand1:
fixes a typo in the docs


mattschouten:
Excellent catch, @maxluchterhand1!

As before, I noticed the same typo shows up in several other version folders (e.g., `.../terraform/v1.12.x/...`, `.../terraform/v1.13x/...`).  Those [versioned folders](https://github.com/hashicorp/web-unified-docs/blob/main/CONTRIBUTING.md#versioned-content) are a quirk of our repository structure.

Would you be willing to update those other occurrences as well, as part of this PR?

PR #1774, opened 2026-02-04
[VAULT] Run Vault on IBM z/OS Container Extensions

schavis:
Create a new deployment guide for running a Vault cluster with HAProxy

attn: @Shobhit-IBM , @yhyakuna 

🔍 [Deploy preview](https://unified-docs-frontend-preview-2y2jizfet-hashicorp.vercel.app/vault/docs/deploy/run-as-zcx-cluster)


content/vault/v1.21.x/content/docs/deploy/run-as-zcx-cluster.mdx
```diff
@@ -0,0 +1,317 @@
...
+  ```shell-session
+  $ docker volume ls | grep vault
+  ```
+
+Volume mount points:
+
+- **`vault-data`**: `/vault/data` (stores Raft snapshots, encrypted storage backend)
+- **`vault-config`**: `/vault/config.d` (HCL configuration files)
+
+## Step 4: Vault configuration file setup
```
brianshumate: **Step 3** heading appears to be missing or skipped?

content/vault/v1.21.x/content/docs/deploy/run-as-zcx-cluster.mdx
```diff
@@ -0,0 +1,317 @@
...
+
+- Simpler configuration: No need to manage certificates on HAProxy or re-encrypt traffic.
+
+- Leader-aware routing: Vault handles leader redirects natively, so the load balancer doesn’t need to interpret HTTP or Vault protocols.
+
+- Better performance: Forwarding raw TCP reduces overhead, giving lower latency and higher throughput.
+
+### Configuration
+
+```plaintext
```
brianshumate: Consider writing the configuration out to `haproxy.cfg` with `cat` or `tee` here, or otherwise mention that the file contents shown need to be written to that file before proceeding.

content/vault/v1.21.x/content/docs/deploy/run-as-zcx-cluster.mdx
```diff
@@ -0,0 +1,317 @@
...
+1. Set your Vault Enterprise license (export before running).
+
+  ```shell-session
+  $ export VAULT_LICENSE="02MV4UU43BK5HGYYTOJZWFQMTMNNEWU33JJVVGC..." 
+  ```
+
+1. Deploying the Vault Enterprise container on zCX.
+
+  ```shell-session
+  $ docker run -d \
```
brianshumate: Unless I'm mistaken, the goal of this guide is to deploy 3 Vault nodes, but there's only the one `docker run` invocation shown here.

This will not achieve the guide's stated goal.

For one, the published ports to localhost (`-p 8200:8200`) cannot be specified more than once (collision), so startup of the other 2 nodes will fail without using different published ports for the other two nodes, or not publishing the ports to localhost at all. The `docker run` examples are missing for the other 2 nodes, however.

Docker Compose would also likely be a better approach to successfully orchestrate deployment a cluster of nodes than trying to use 3 `docker run` commands.

There is also the issue of a folder structure for the 3 nodes such that the volumes for configuration and data work as expected for all 3 nodes. Again, Compose would likely help with this. You could also consider the Terraform Docker Provider.

PR #1872, opened 2026-02-19
[VAULT-42462] Update docs to include `fetch_groups` and `distributed_claim_access_token` for Azure AD Auth

abbyck:
https://hashicorp.atlassian.net/browse/VAULT-42462

### Description:

Adds documentation for two new parameters that support fetching group memberships directly from the Microsoft Graph API when using the Azure (Entra ID) provider with the JWT/OIDC auth method.

#### Changes

[jwt.mdx]: Add the distributed_claim_access_token parameter to the JWT login endpoint. This access token is required when fetch_groups is enabled and allows Vault to authenticate to the Microsoft Graph API to retrieve group memberships.
 Document the fetch_groups provider config option under the "Optional azure-specific configuration" section. This boolean parameter lets Vault fetch groups through GET /me/memberOf instead of relying on the groups claim in the token, bypassing both the 200-group token limit and the per-application 1,000 group assignment limit in Microsoft Entra ID.


content/vault/v2.x (rc)/content/docs/auth/jwt/oidc-providers/azuread.mdx
```diff
@@ -155,6 +155,18 @@ To set the proper permissions on the Azure app:
    EOH
    ```
 
+   You can include the following optional parameter in the `provider_config` object:
+
+   - `fetch_groups` `(bool: false)` - If set to `true`, Vault fetches groups directly
+     from the Microsoft Graph API through `GET /me/memberOf` instead of relying on the
```
roh-a: I think we need to mention `GET /me/getMemberObjects` instead of `GET /me/memberOf`. 
Let's also add a link to the official Microsoft docs for this API.

stuti-sr (approved):
provided nit comments. 
Rest looks good to me!!

PR #1912, opened 2026-02-27
Enable Incremental Builds Beta Test

RubenSandwich:
## Relevant links

* [🎟️ Detect changed files and build a changedFiles.json metadata file](https://ibm.monday.com/boards/18401913557/pulses/10906807460)
* [🎟️ Allow prebuild to run only on the files in changedFiles.json](https://ibm.monday.com/boards/18401913557/pulses/10906804910)
* [🎟️ Modify APIs to serve prod content for files not changed](https://ibm.monday.com/boards/18401913557/pulses/10906805144)

## Description

Implements incremental builds; through a layered "CDN approach" where only changed assets are loaded in the PR Build's CDN and the rest are served from the Production CDN.

This is also gated to using a "inc-build" label to turn it on during PR Builds so that we can beta test and make sure that everything is working before incremental builds becomes the default strategy for PR Builds.

## Testing

* Testing has been provided for the following two main new functions:
    * [getFilesUsingPartial](scripts/utils/get-files-using-partial.test.mjs)
    * [fetchFile](app/utils/file.test.ts)
* Full functionality can be tested in this PR: https://github.com/hashicorp/web-unified-docs/pull/1984


scripts/utils/get-files-using-partial.mjs
```diff
@@ -0,0 +1,127 @@
...
+	// Pattern: content/{product}/global/partials/{rel-path}
+	// Include: a relative path (single or double quoted) whose suffix is
+	// `global/partials/{rel-path}`. Quote-agnostic because we match the substring
+	// between the quotes, not the quotes themselves.
+	if (parts[2] === 'global' && parts[3] === 'partials') {
+		const relPath = parts.slice(4).join('/')
+		const includeSubstring = `global/partials/${relPath}`
+		const allFiles = walkMdxFiles(`content/${product}`)
+
+		return filesContaining(allFiles, includeSubstring)
```
LeahMarieBush: Small nitpick- I think it would be more consistent to call `filesContaining` the same way for each scenario. Scenario 1 and 3 have the `@include` prefix but scenario 2 does not.

PR #1918, opened 2026-03-01
Fix a typo

Wajktor13:


Wajktor13:
> @Wajktor13, thanks for this fix! It looks like this typo was introduced in 1.14, but also exists in 1.15 alpha docs. Would you mind fixing it there as well? If not, no worries I'll make sure it gets updated. Thanks again!
> 
> https://github.com/hashicorp/web-unified-docs/blob/main/content/terraform/v1.15.x%20(alpha)/docs/language/resources/index.mdx

done

crw:
Thank you!

PR #1983, opened 2026-03-11
Documentation of the Common Criteria Flag

kitography:
This PR adds documentation for the feature flag common_criteria_mode which addresses some of the incompatibilities between the NIAP common criteria requirements and Vault APIs that need to be addressed ahead of getting NIAP certification.


content/vault/v2.x (rc)/content/api-docs/secret/pki/common-criteria.mdx
```diff
@@ -0,0 +1,85 @@
...
+---
+
+~> **Note**: This feature flag is being introduced before starting NIAP
+   certification in order to address certain compatibility differences between
+   Vault behaviour and the Common Criteria Requirements.  Vault PKI does not yet
+   have NIAP certification, and in the future, enabling this feature flag will be
+   necessary but not sufficient to run Vault in a manner compliant with common
+   criteria.
+
+# Common Criteria Flag
```
schavis: I'm very confused about the purpose of this page.

Just to confirm, this is a configuration parameter and an [activation flag](https://developer.hashicorp.com/vault/api-docs/system/activation-flags)?

If it's a configuration parameter, why are we documenting how to enable it and the limitations/affects of the configuration setting in the API docs?

Generally speaking, content related to enabling a configuration parameter (and the effect it has on your server) should be in the configuration section. The API docs should focus on the available API endpoints, what they do, and what the request/response objects look like.

From skimming the rest of the PR, it doesn't seem like this is in the right place.

schavis (changes requested):

schavis:
This PR appears abandoned. I'm going to close it if only for the overwhelming number of conflicts. Feel free to recreate the PR against main

PR #2036, opened 2026-03-19
Forward-porting feature

RubenSandwich:
## 🔗 Relevant links

- [Preview link](https://unified-docs-frontend-preview-hrtf4lwjy-hashicorp.vercel.app/) 🔎
- [Monday task](https://ibm.monday.com/boards/18400959349) 🎟️

## 🗒️ What

Adds an automated forward-port GitHub Actions workflow. When a versioned docs PR is merged with a `forward-port:<slug>` label, the workflow automatically opens a new PR that copies the changed files from the source version directory into the target version directory (e.g. `content/terraform/v1.14.x/` → `content/terraform/v1.15.x/`).

Also supports manual triggering via `workflow_dispatch` with a PR number & required inputs.

Split out the big steps of this into node scripts so it would be easier to implement and add testing for. To describe what happens in gha- first, after gathering the information from the PR- the target information is resolved (either grabbed from the config file or PR comment) and saved to github env variables. Next the changed files from the PR are determined and saved to a temp file to avoid losing the changes on branch switch. The changes are forward-ported and applied only to the files relevant to the product. 

> This is because a common use-case may be running this against repo-sync PRs, and since there are tons of changes from many different products there, it has to sift through it simply and rigidly. 

After that the forward port PR is created and a comment is posted back to the source PR

I also added in a [readme](https://github.com/hashicorp/web-unified-docs/blob/09e83af37924ec308b1afa1a4296539f1c026323/scripts/forward-port/forward-port-README.md) for this to make it easier for people to use and be aware of edge cases. Keeping integration testing steps there for us in the future.

## 🤷 Why

When a docs change lands in one version directory (e.g. `v1.14.x`), the same change typically needs to land in the next version too. Before that process was done locally with a script- this workflow attempts to automate that

In this PR you may notice that info is often explicitly requested from the user, is repeated, or that a forward-port label always has to be present on the PR, even when doing the manual dispatch with the JSON override (it doesn't really need the label to run). This is there just so that each PR that has the forward port feature will ALWAYS be in a consistent state, no matter how the workflow was ran. This was to save me some mental load and not worry about potential weird edge cases. It's rigid by design. For a pilot, this is fine. Later on when this is working well, the rigidness can be eased up (i.e. allowing multiple forward port labels on a PR, etc)

## 🛠️ How

**Automatic path:** A merged PR with a `forward-port:<slug>` label triggers the workflow. The slug is looked up in `forward-port-config.yml` to get `sourceVersionFolder`, `targetVersionFolder`, `targetProduct`, and `targetBranch`. For sensitive/unreleased content where the routing shouldn't be public in the config, a `/forward-port` PR comment can supply the routing instead

**Manual path:** `workflow_dispatch` with a `prNumber` + either `labelSlug` (config lookup) or `overrideJson` (raw routing fields). Note: `workflow_dispatch` requires the workflow to be on the default branch before GitHub exposes the UI trigger, so this will have to be tested once merged sadly. However I tested the code as much as I could locally, so fingers crossed.

**Key implementation detail:** The workflow does two `actions/checkout` calls- the first at the merge commit (to access scripts + source content), the second at the target branch. Right now some target branches may not have this script, so the gha runner creates a copy just to have access to that code which is only pretty much doing text manipulation.

The config is always fetched from `main` at run time via the GitHub API, not from the triggering PR's branch

## 🧪 Testing

Here I'll include steps to test this manually with PRs I've already set up and also instructions to run the integration tests I made. These integration tests run destructive git commands on test repos that are initialized at the start of the tests, so run these only if you're comfortable with that. Otherwise you can test the functionality manually with these steps:

## Reading from config file:

1. create a branch off `rn-forward-porting-test-ga`

1. Make a change to docs in: `content/terraform-forward-port-testing` , commit it, push it 

1. create a PR and set the merge base to `rn-forward-porting-test-ga`

1. add this label that coincides with a label in the config file: `forward-port:tf-forward-porting-testing`

1. ***REMEMBER TO CHANGE YOUR BASE SO THAT IT MERGES INTO: rn-forward-porting-test-ga, BY DEFAULT PRS MERGE INTO MAIN, YOU DO NOT WANT THAT!***

1. merge the PR

1. The forward porting workflow will open a PR into: [rn-forward-porting-test-rc](https://github.com/hashicorp/web-unified-docs/pull/2460/changes#diff-1d8bceee86194a808badbe791f573b6ba31ac1950f8cc83a6e42fb3bbe2c42dfR78)

1. You can check workflow runs here: [forward-port-pr.yml](https://github.com/hashicorp/web-unified-docs/actions/workflows/forward-port-pr.yml)

1. look at the comments on the merged PR and verify that it links to the forward port pr

1. verify the content in the forward port pr

## Reading from PR comment (when the config file cant contain information on embargoed content):

1. create a branch off `rn-forward-porting-test-ga`

1. Make a change to docs in: `content/terraform-forward-port-testing` , commit it, push it 

1. create a PR and set the merge base to `rn-forward-porting-test-ga`

1. add a forwardport label that *DOES NOT* coincide with a label in the config file: i.e. `forward-port:my-secret-test-slug`

1. add a PR comment as such
```
/forward-port forward-port:my-secret-test-slug
sourceVersionFolder: v1.14.x
targetProduct: terraform-forward-port-testing
targetBranch: rn-forward-porting-test-rc
targetVersionFolder: v1.15.x
```
6. ***REMEMBER TO CHANGE YOUR BASE SO THAT IT MERGES INTO: rn-forward-porting-test-ga, BY DEFAULT PRS MERGE INTO MAIN, YOU DO NOT WANT THAT!***
 
7. merge it
 
8. The forward porting workflow will open a PR into: [rn-forward-porting-test-rc](https://github.com/hashicorp/web-unified-docs/pull/2460/changes#diff-1d8bceee86194a808badbe791f573b6ba31ac1950f8cc83a6e42fb3bbe2c42dfR78)
 
9. look at the comments on the merged PR and verify that it links to the forward port pr
 
10. verify the content in the forward port pr

## If you want to run the integration tests:

```bash
# Unit tests
npx vitest run scripts/forward-port/resolve-target.test.ts

# Integration — apply changes
cp scripts/forward-port/apply-forward-port-changes.integration.ts \
   scripts/forward-port/apply-forward-port-changes.integration.test.ts
npx vitest run scripts/forward-port/apply-forward-port-changes.integration.test.ts
rm scripts/forward-port/apply-forward-port-changes.integration.test.ts

# Integration — get changed files
cp scripts/utils/get-changed-content-files-forward-port.integration.ts \
   scripts/utils/get-changed-content-files-forward-port.integration.test.ts
npx vitest run scripts/utils/get-changed-content-files-forward-port.integration.test.ts
rm scripts/utils/get-changed-content-files-forward-port.integration.test.ts

# E2E pipeline
cp scripts/forward-port/forward-port.e2e.integration.ts \
   scripts/forward-port/forward-port.e2e.integration.test.ts
npx vitest run scripts/forward-port/forward-port.e2e.integration.test.ts
rm scripts/forward-port/forward-port.e2e.integration.test.ts

```

## 💭 Anything else?

**Before this goes to production, three things need updating in the workflow:**
1. `CONFIG_BRANCH` is hardcoded to `rn-forward-porting-test-ga` in both routing steps, change to `main`
3. `pull_request.branches` trigger is hardcoded to `rn-forward-porting-test-ga`,  change to `main`
4. Most entries in `forward-port-config.yml` have placeholder version/branch values 

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.
- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
- [x] If applicable, I've documented the impact of any changes to security controls.


.github/forward-port-config.yml
```diff
@@ -0,0 +1,75 @@
+Boundary:
+  targetBranch: TF-next
+  targetVersion: v1.1.0
```
RubenSandwich: Rename to targetVersionFolder

williamdalessandro:
updated the original comment- putting ANOTHER personal reminder to change the trigger and config branches

.github/workflows/forward-port-pr.yml
```diff
@@ -0,0 +1,354 @@
...
+      # SOURCE_VERSION_FOLDER comes from resolve-target (PR event) or workflow_dispatch inputs.
+      # Source files were saved to /tmp before the target-branch checkout.
+      - name: Apply forward-port changes
+        run: |
+          if [ -z "${SOURCE_VERSION_FOLDER}" ]; then
+            echo "::error::SOURCE_VERSION_FOLDER is not set — check resolve-target step or workflow_dispatch inputs"
+            gh pr comment "${SOURCE_PR_NUMBER}" \
+              --body "@${SOURCE_PR_AUTHOR} :x: Forward-port failed: \`SOURCE_VERSION_FOLDER\` was not set. This is a bug — please file an issue." \
+              --create-if-none \
+              --edit-last || true
```
RubenSandwich: Why is this OR'ed with true?

scripts/forward-port/apply-forward-port-changes.mjs
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Rewrites the source-version segment of a file path to the target version.
+ * e.g. content/terraform/v1.14.x/docs/file.mdx → content/terraform/v1.15.x/docs/file.mdx
+ */
+function toTargetPath(srcPath) {
+	return srcPath.replace(sourceVersion, targetVersion)
+}
+
+// filter out only the targetProduct files/directories from the changesFile
```
RubenSandwich, line 42: Good catch here to make sure we are only doing this product by product, because changeFiles could contain changes from two products.

scripts/forward-port/forward-port-README.md
```diff
@@ -0,0 +1,176 @@
...
+| `targetBranch` | The branch the forward-port PR will be opened against |
+| `targetVersionFolder` | The version directory to copy files into (e.g. `v1.20.x`) |
+
+All four fields are required. To add a new forward-port route, add an entry to this file and get it merged into `main` before merging your content PR — the workflow reads the config from `main` at run time, not from your branch.
+
+> **Important:** The workflow always reads the config from `main` (not from your PR branch). Your config change must be merged into `main` before the workflow will see it. If you merge your content PR before the config entry lands in `main`, the workflow will fail to find the slug.
+
+---
+
+## Comment fallback (Scenario B)
```
RubenSandwich, line 63: Interesting. I guess that the authors asked for this. Kind of makes sense. No comment on it overall, just an interesting use case.

scripts/forward-port/forward-port-README.md
```diff
@@ -0,0 +1,176 @@
...
+
+---
+
+## What files get ported
+
+Only files under `content/<targetProduct>/` are ported. Files from other products in the same PR are automatically excluded. The version segment in each file path is rewritten from `sourceVersionFolder` to `targetVersionFolder`.
+
+- **Added / modified** files are copied from the source version directory to the target version directory.
+- **Deleted** files are deleted from the target version directory.
+- Partial files (`/partials/`) are excluded — only directly-changed files are ported.
```
RubenSandwich: We should port over partial files, and the code looks it does do that: https://github.com/hashicorp/web-unified-docs/pull/2036/changes#diff-f5d771c1bdaf58c49f827d985aacea8ea59081ecdf32cfa2911572b5cf51805cR79.

So this line should be removed. Partial files are included, we just don't auto include files that that partial file applies to. We need that information during incremental builds, but not here.

scripts/forward-port/resolve-target.mjs
```diff
@@ -0,0 +1,244 @@
...
+
+	matched = parseForwardPortComment(commentText, commentFile, slug)
+	console.log(
+		`Using /forward-port comment → targetBranch=${matched.targetBranch}, ` +
+			`sourceVersionFolder=${matched.sourceVersionFolder}, targetVersionFolder=${matched.targetVersionFolder}, ` +
+			`targetProduct=${matched.targetProduct}`,
+	)
+}
+
+// --- Write to GITHUB_ENV ---
```
RubenSandwich: @williamdalessandro Instead of writing this directly to the GITHUB_ENV in this script, we should:
1. Have this script return the needed values as JSON
2. Have the output'ed JSON from 1 written to the GITHUB_ENV in using shell scripting: https://github.com/hashicorp/web-unified-docs/pull/2036/changes#diff-faf9cb6ecdecbe1c8a3d00bb95f1415a51d40397b75a9933bd093c86d7a1753fR141

This script shouldn't change the machine it runs on, unless it absolutely need to and I don't believe it does. We can do it outside of the script, plus we should be more explicit when we write things to GITHUB_ENV.

This will also clean up the test file for this script. Because right now the test file is using "exec", which raises some red flags for me.

RubenSandwich (commented):
I believe we can simplify resolve-targets.

scripts/utils/get-changed-content-files-cli.mjs
```diff
@@ -0,0 +1,24 @@
+/**
+ * Copyright IBM Corp. 2024, 2026
```
RubenSandwich: I don't think this needs to exist as it's own file. You can detect a script is being run through the command line or if it's just being imported as a module like so:

```
/**
 * Returns true when this module is executed directly by Node, e.g.
 * `node scripts/utils/get-changed-content-files.mjs`.
 */
function isRunFromCommandLine() {
	if (!process.argv[1]) {
		return false
	}

	return import.meta.url === pathToFileURL(process.argv[1]).href
}

if (isRunFromCommandLine()) {
	void getChangedContentFiles()
}
```

From there you can get the command line args you need:
1. mergeBase
2. includePartials
3. output

That would remove this file and it's added tests.

.github/workflows/forward-port-pr.yml
williamdalessandro: mostly kept it there just as defensive code so that if theres an issue with the actual gh command run it won't stop the workflow and we can fallback to our own error controls. definitely is a bit redundant because I'm mostly just doing an exit 1 after most of these, but for the [one situation where it isnt the case](https://github.com/hashicorp/web-unified-docs/pull/2036/changes/BASE..5921c59ee67dab7edcfa64cea41d6e16ad7e035f#diff-faf9cb6ecdecbe1c8a3d00bb95f1415a51d40397b75a9933bd093c86d7a1753fR313-R314), it lets us continue. I'm not married to the pattern though. 

But while we're looking at that one situation I mentioned (it's this edge case about what to do when there's nothing to forward-port) its something I'm gonna fix up. I'm going to change this and exit the workflow after the comment is posted saying "there are no changes to forward-port". Doesn't make much sense to continue with the workflow after looking at this again

I also feel like removing the --edit-last flag on these comments so that it doesn't override any other comments that may already exist there (mostly didnt think too deep about this because this is being ran on PRs that have already been merged)

scripts/forward-port/forward-port-README.md
williamdalessandro: I included this line because the exact reason you said- the language used here with "/partials/ are excluded" is just misleading. I'll remove that line and replace it to be more clear

scripts/forward-port/resolve-target.mjs
williamdalessandro: I'm ok with making this change

scripts/utils/get-changed-content-files-cli.mjs
williamdalessandro: Thank you for the suggestion!

I split it out because the way I had it before was causing the compile prebuild/prebuild to break and this was the quick fix. I'll change this to eliminate the extra file. the tests I'll keep/update though because it is nice to have checks with the new usage

williamdalessandro (commented):
Thank you for the feedback @RubenSandwich! Going modify this accordingly

scripts/utils/get-changed-content-files-cli.mjs
williamdalessandro: ~~Hey I implemented your suggestion but still ran into compile-prebuild issues again. This time I dug deeper into the issue and it has to do with the `bun build --bytecode` compile step. Adding `import.meta.url` to the file makes it runtime-only since its value as the file's URL on disk is determined at the moment of execution and it breaks the bun compiler with `error: Failed to generate bytecode for ./prebuild.js`~~

~~I could either keep the file separated as is or hardcode the URL during the check like this~~

```
function isRunFromCommandLine() {
  return process.argv[1]?.endsWith('get-changed-content-files.mjs') ?? false
}
```
~~However it's a bit fragile and can break if the file gets renamed~~

LeahMarieBush (commented):
Nice job! I like how comprehensive the testing is for this and the clear documentation.
I have one question related to the PRs that are opened after the action is completed. I was looking at some of the PRs for testing and noticed that the account opening the PR is github-actions which does not have CLA access for this repo. Should this be changed to hashibot so it has proper permissions?

williamdalessandro:
> Nice job! I like how comprehensive the testing is for this and the clear documentation. I have one question related to the PRs that are opened after the action is completed. I was looking at some of the PRs for testing and noticed that the account opening the PR is github-actions which does not have CLA access for this repo. Should this be changed to hashibot so it has proper permissions?

@LeahMarieBush thank you for reminding me about this! I was wondering the same thing at a certain point but forgot to follow through- it should be easy enough to modify

williamdalessandro:
> Nice job! I like how comprehensive the testing is for this and the clear documentation. I have one question related to the PRs that are opened after the action is completed. I was looking at some of the PRs for testing and noticed that the account opening the PR is github-actions which does not have CLA access for this repo. Should this be changed to hashibot so it has proper permissions?

@LeahMarieBush @RubenSandwich Changed this up to use the hashibot - can see a successful forward port [here](https://github.com/hashicorp/web-unified-docs/pull/2541) . The added the full-build tag and the preview worked, so I'm pretty sure the build preview issues Ive been seeing on these forward-port PRs are related to the issue brought in [this support ticket](https://ibm-hashicorp.slack.com/archives/C09LU1THDDE/p1780326315168079?thread_ts=1779486203.100629&cid=C09LU1THDDE)

LeahMarieBush (approved):
LGTM!

scripts/utils/get-changed-content-files-cli.mjs
williamdalessandro: Updating here for posterity's sake- was able to implement the original suggestion after @RubenSandwich pointed out that the newer versions of bun support `import.meta.url`, so disregard my previous comment

williamdalessandro (approved):

PR #2076, opened 2026-03-24
prevent root escape in @include

mcollao-hc:

### Terraform

### DIFF
--- a/scripts/prebuild/mdx-transforms/include-partials/remark-include-partials.mjs
+++ b/scripts/prebuild/mdx-transforms/include-partials/remark-include-partials.mjs
@@ -31,9 +31,56 @@ const isPathWithinRoot = (rootPath, candidatePath) => {
 	)
 }
 
-const normalizeIncludePath = (includePath) => {
-	if (includePath.startsWith('/')) {
-		return `.${includePath}`
+const stripLeadingSlash = (includePath) => {
+	return includePath.replace(/^\/+/, '')
+}
+
+const getCanonicalPath = (targetPath) => {
+	return fs.realpathSync(targetPath)
+}
+
+const resolveIncludeFilePath = ({
+	rawPath,
+	partialsDir,
+	productGlobalPartialsDir,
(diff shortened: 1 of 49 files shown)


scripts/prebuild/mdx-transforms/include-partials/remark-include-partials.mjs
```diff
@@ -22,6 +22,74 @@ export const PARTIALS_ALIAS = {
...
+	const requestedPath = stripLeadingSlash(
+		isGlobalAlias ? rawPath.slice(globalPrefix.length) : rawPath,
+	)
+	const includeBaseDir = isGlobalAlias ? topLevelPartialsDir : partialsDir
+	const allowedRoots = (
+		isGlobalAlias
+			? [topLevelPartialsDir]
+			: [partialsDir, productGlobalPartialsDir]
+	)
+		.filter(Boolean)
```
rmainwork: I noticed we're using `.filter(Boolean)` here. If we're mainly guarding against `productGlobalPartialsDir` being optional/undefined, would it be worth explicitly checking for that above so the intent is clearer to future readers?

rmainwork (changes requested):
Hey! Thanks for getting this PR opened. I like the direction, but we need to tweak a few things before we can merge:

- To keep the security review super focused, let's pull the ESLint/style changes out. (Happy to merge those in a follow-up PR though!)
- We need to keep the ability to do cross-product includes. I know it seems like something we'd want to lock down, but we're actually relying on that exact behavior for a feature coming down the pipeline.
- For the new wrapper functions, let's just use the native Node functions directly where possible so we don't add extra layers of abstraction.

Thanks again for grabbing this. Let me know when it's ready for another look!

PS:  Just out of curiosity, were you using Copilot or Cursor for some of these refactors? I only ask because I've noticed those tools have a strong habit of adding unnecessary wrapper functions and locking down business logic a little too aggressively, and it looks exactly like the kind of things it tries to sneak into my PRs too! 😂

PR #2126, opened 2026-04-02
[VAULT] Update the LTS docs to discuss SC2 instead

schavis:


content/vault/v2.x (rc)/content/docs/enterprise/lts.mdx
```diff
@@ -1,177 +1,62 @@
...
+for more detail.
+
+Product update            | Base support | Extended support | Sustained support 
+------------------------- | ------------ | ---------------- | -----------------
+New features              | Yes          | No               | No 
+Non-critical bug fixes    | Yes          | No               | No 
+Product improvements      | Yes          | No               | No
+Critical bug fixes        | Yes          | Yes              | No
+Security patches          | Yes          | Limited          | No
+
```
jonathanfrappier: Worth adding something for access to tech support so sustained support doesn't seem useless? Though not exactly clear from their support doc what that means - can i open support cases? or just access old support cases.
```suggestion
Security patches          | Yes          | Limited          | No
Technical Support Case Access | Yes          | Yes          | Yes
```

PR #2216, opened 2026-04-13
fix: parameters for ListResourceConfigSchema

thiagoarrais:
This changes the parameter types for the `ListResourceConfigSchema` example method to match the ones documented on the [ListResource interface](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/list#ListResource)


SBGoods (approved):
Thank you for the contribution @thiagoarrais

PR #2231, opened 2026-04-14
Patch prebuild script

rmainwork:
The MDX transform was compiling two sets of versions into the paths of links. This is incorrect and was causing broken links. Also centralizes the version in path detection to one utility function for easier refactoring and updating in the future.


RubenSandwich (approved):
This looks great! No notes. 🚀

PR #2244, opened 2026-04-15
fix ruby comments from using "//" to "#" and use correct "delete" method for the erase call

danabucci:
This token-helper ruby script works well, but ruby comments are `#` not `//`. This change simply replaces the comment marker with the correct one for ruby.

Additionally, this pull request replaces the nonexistent `delete!` method for hashes with the correct `delete`


schavis (approved):

schavis:
@danabucci Thanks for the fix!

PR #2248, opened 2026-04-15
add known issue for ent license on `openldap` alias

vinay-gopalan:
This PR adds a known issue where the new v2.x Enterprise feature — self-managed static roles for LDAP — is not enabled when using the `openldap` plugin alias due to enterprise license checking.

The workaround is described in the PR. Users can instead use the `ldap` plugin instead of the `openldap` alias. This is the same plugin under the hood and has the feature enabled

### DIFF
--- a/content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
+++ b/content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
@@ -1,4 +1,4 @@
-### LDAP secrets self-managed static roles feature is not enabled on the `openldap` built-in alias
+### LDAP secrets self-managed static roles feature is not enabled on the `openldap` built-in type alias
 
 <!-- BEGIN: Vault:=v2.x -->
 
@@ -10,8 +10,8 @@
 
 #### Issue
 
-Self-managed static roles feature is not enabled on the `openldap` built-in alias for the LDAP Secrets plugin.
-Trying to use the feature with the `openldap` plugin results in the following error:
+Self-managed static roles feature is not enabled when using the `openldap` built-in TYPE alias for the LDAP Secrets plugin.
+Trying to use the feature with the `openldap` type results in the following error:
 
 ```
 * 1 error occurred:
@@ -20,7 +20,7 @@ Trying to use the feature with the `openldap` plugin results in the following er
 
 #### Workaround
 
-Use the `ldap` plugin to enable the new self-managed mounts. Enterprise license checking enables the feature on `ldap`:
+Use the `ldap` plugin type to enable the new self-managed mounts. Enterprise license checking enables the feature on the mount:
 
 ```
 vault secrets enable -path=<PATH> ldap
--- a/content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
+++ b/content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
@@ -1,31 +1,32 @@
-### LDAP secrets self-managed static roles feature is not enabled on the `openldap` built-in type alias
+### LDAP secrets self-managed static roles feature does not work with the `openldap` built-in type alias ((#self-managed-openldap-fails)) <EnterpriseAlert inline="true" />
 
 <!-- BEGIN: Vault:=v2.x -->
 
 | Change      | Status   | Vault edition | Affected version | Fixed version
 | ----------- | -------- | ------------- | ---------------- | -------------
-| Known issue | Closed   | Enterprise    | 2.0.0            | 2.0.1
+| Known issue | Closed   | Enterprise    | 2.0.0            | No
 
 <!-- END: Vault:=v2.x -->
 
 #### Issue
 
-Self-managed static roles feature is not enabled when using the `openldap` built-in TYPE alias for the LDAP Secrets plugin.
-Trying to use the feature with the `openldap` type results in the following error:
+Self-managed static roles feature does not work when using the `openldap` built-in TYPE alias for the LDAP Secrets plugin.
+Using the `openldap` built-in TYPE alias for the LDAP Secrets plugin with self-managed static roles will return the following error:
 
-```
+```text
 * 1 error occurred:
         * self-managed static roles feature requires a valid Vault Enterprise license
 ```
 
 #### Workaround
 
-Use the `ldap` plugin type to enable the new self-managed mounts. Enterprise license checking enables the feature on the mount:
+Use the `ldap` plugin type to configure the new self-managed mounts:
 
 ```
-vault secrets enable -path=<PATH> ldap
+vault secrets enable -path=<mount_path> ldap
 
-vault write <PATH>/config \
-    ...
+vault write <mount_path>/config \
     self_managed=true
 ```
+
+Vault will verify your enterprise license and enable the feature for the mount.
--- a/content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
+++ b/content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
@@ -1,4 +1,4 @@
-### LDAP secrets self-managed static roles feature does not work with the `openldap` built-in type alias ((#self-managed-openldap-fails)) <EnterpriseAlert inline="true" />
+### LDAP plugins must use `ldap` type for self-managed static roles ((#self-managed-openldap-fails)) <EnterpriseAlert inline="true" />
 
 <!-- BEGIN: Vault:=v2.x -->
 
@@ -10,8 +10,8 @@
 
 #### Issue
 
-Self-managed static roles feature does not work when using the `openldap` built-in TYPE alias for the LDAP Secrets plugin.
-Using the `openldap` built-in TYPE alias for the LDAP Secrets plugin with self-managed static roles will return the following error:
+Self-managed static roles do not work when using the `openldap` built-in `TYPE` alias for the LDAP Secrets plugin.
+Enabling a new LDAP mount with the `openldap` built-in TYPE alias with self-managed static roles returns the following error:
 
 ```text
 * 1 error occurred:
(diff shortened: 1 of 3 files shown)


content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,31 @@
...
+| Change      | Status   | Vault edition | Affected version | Fixed version
+| ----------- | -------- | ------------- | ---------------- | -------------
+| Known issue | Closed   | Enterprise    | 2.0.0            | 2.0.1
+
+<!-- END: Vault:=v2.x -->
+
+#### Issue
+
+Self-managed static roles feature is not enabled on the `openldap` built-in alias for the LDAP Secrets plugin.
+Trying to use the feature with the `openldap` plugin results in the following error:
```
fairclothjm:
```suggestion
Self-managed static roles feature is not enabled when using the `openldap` built-in TYPE alias for the LDAP Secrets plugin.
Trying to use the feature with the `openldap` type results in the following error:
```

fairclothjm: this is documented as the `type` so we should probably use that language:

```
$ vault secrets enable --help
Usage: vault secrets enable [options] TYPE

  Enables a secrets engine. By default, secrets engines are enabled at the path
  corresponding to their TYPE, but users can customize the path using the
  -path option.
```

See also
- https://developer.hashicorp.com/vault/docs/commands/secrets/enable
- https://developer.hashicorp.com/vault/api-docs/system/mounts#enable-secrets-engine

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+```
+* 1 error occurred:
+        * self-managed static roles feature requires a valid Vault Enterprise license
+```
+
+#### Workaround
+
+Use the `ldap` plugin to enable the new self-managed mounts. Enterprise license checking enables the feature on `ldap`:
```
fairclothjm:
```suggestion
Use the `ldap` plugin type to enable the new self-managed mounts. Enterprise license checking enables the feature on the mount:
```

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
vinay-gopalan: Good call! Updated

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,31 @@
+### LDAP secrets self-managed static roles feature is not enabled on the `openldap` built-in type alias
+
+<!-- BEGIN: Vault:=v2.x -->
+
+| Change      | Status   | Vault edition | Affected version | Fixed version
+| ----------- | -------- | ------------- | ---------------- | -------------
+| Known issue | Closed   | Enterprise    | 2.0.0            | 2.0.1
```
schavis:
```suggestion
| Known issue | Closed   | Enterprise    | 2.0.0            | No
```
We don't promise things for future versions (just in case things change/slip)

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+<!-- END: Vault:=v2.x -->
+
+#### Issue
+
+Self-managed static roles feature is not enabled when using the `openldap` built-in TYPE alias for the LDAP Secrets plugin.
+Trying to use the feature with the `openldap` type results in the following error:
+
+```
```
schavis:
````suggestion
```text
````

File: content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx [hunk 5]
```diff
@@ -0,0 +1,31 @@
...
+
+Use the `ldap` plugin type to enable the new self-managed mounts. Enterprise license checking enables the feature on the mount:
+
+```
+vault secrets enable -path=<PATH> ldap
+
+vault write <PATH>/config \
+    ...
+    self_managed=true
+```
```
schavis:
````suggestion

Configure the mount for self-managed static roles:

```shell-session
$ vault write <mount_path>/config \
    self_managed=true
```

Vault will verify your enterprise license and enable the feature for the mount.
````

Why are we showing the enable step as part of the workaround? Do they have to create a whole new plugin instance?

File: content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx [hunk 6]
```diff
@@ -0,0 +1,31 @@
...
+| Change      | Status   | Vault edition | Affected version | Fixed version
+| ----------- | -------- | ------------- | ---------------- | -------------
+| Known issue | Closed   | Enterprise    | 2.0.0            | 2.0.1
+
+<!-- END: Vault:=v2.x -->
+
+#### Issue
+
+Self-managed static roles feature is not enabled when using the `openldap` built-in TYPE alias for the LDAP Secrets plugin.
+Trying to use the feature with the `openldap` type results in the following error:
```
schavis:
```suggestion
Using the `openldap` built-in TYPE alias for the LDAP Secrets plugin with
self-managed static roles may return the following error:
```
I think this is what we're saying? The text as written was a little hard to parse.

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,31 @@
+### LDAP secrets self-managed static roles feature is not enabled on the `openldap` built-in type alias
```
schavis:
```suggestion
### LDAP secrets self-managed static roles feature is not enabled by default ((#ldap-self-managed))
```
I think this is the actual problem we're describing? Not that it's not enabled, but that it's not enabled by default?

"You need to enable feature X before using it" doesn't really seem like a known issue in, and of, itself?

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,31 @@
...
+
```
schavis, line 2: Please also add an entry to the summary table file for 2.x: `content/vault/global/partials/important-changes/summary-tables/2xx.mdx`

schavis (changes requested):

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
vinay-gopalan: This feature requires creating a new mount. If the mount is not enabled with the above type, the feature will not work, hence we want to document exactly how to enable the mount to use this feature

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
vinay-gopalan: LDAP Secrets can be enabled in multiple ways. Either by enabling the `ldap` type or by enabling `openldap` type. The problem here is that the self-managed static roles (enterprise-only) feature does not work when using the `openldap` type, so we want to document for users that they should explicitly use the `ldap` type when enabling their secret engine

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
vinay-gopalan: Thanks! Added

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,32 @@
+### LDAP secrets self-managed static roles feature does not work with the `openldap` built-in type alias ((#self-managed-openldap-fails)) <EnterpriseAlert inline="true" />
```
schavis:
```suggestion
### LDAP plugins must use `ldap` type for self-managed static roles ((#self-managed-openldap-fails)) <EnterpriseAlert inline="true" />
```

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,32 @@
...
+| Change      | Status   | Vault edition | Affected version | Fixed version
+| ----------- | -------- | ------------- | ---------------- | -------------
+| Known issue | Closed   | Enterprise    | 2.0.0            | No
+
+<!-- END: Vault:=v2.x -->
+
+#### Issue
+
+Self-managed static roles feature does not work when using the `openldap` built-in TYPE alias for the LDAP Secrets plugin.
+Using the `openldap` built-in TYPE alias for the LDAP Secrets plugin with self-managed static roles will return the following error:
```
schavis:
```suggestion
Self-managed static roles do not work when using the `openldap` built-in `TYPE`
alias for the LDAP Secrets plugin. Enabling a new LDAP mount with the `openldap`
built-in TYPE alias with self-managed static roles returns the following error:
```

content/vault/global/partials/important-changes/known-issue/openldap-license-check.mdx
```diff
@@ -0,0 +1,32 @@
...
+Use the `ldap` plugin type to configure the new self-managed mounts:
+
+```
+vault secrets enable -path=<mount_path> ldap
+
+vault write <mount_path>/config \
+    self_managed=true
+```
+
+Vault will verify your enterprise license and enable the feature for the mount.
```
schavis:
````suggestion

1. Enable a new LDAP plugin for self-managed static roles:

  ```shell-session
  $ vault secrets enable -path=<mount_path> ldap
  ```

1. Use the `ldap` plugin type to configure the new mount:

  ```shell-session
  $ vault write <mount_path>/config \
      self_managed=true
  ```

When you set `self_managed` to `true`, Vault verifies your enterprise license
and enables the feature for the new mount.
````

schavis (approved):

PR #2261, opened 2026-04-16
[WIP] Update OS secret engine to add HCL configuration examples

drewmullen:
Do not merge until after [TFVP resources](https://github.com/hashicorp/terraform-provider-vault/pull/2865) are merged

### Terraform


content/vault/v2.x/content/docs/secrets/os.mdx
```diff
@@ -32,31 +32,43 @@ Most secrets engines must be configured in advance before they can perform their
...
-    ```
+<Tabs>
 
-2. Enable the OS secrets engine:
-    ```shell
-    $ vault secrets enable -path=os vault-plugin-secrets-os
-    ```
+<Tab heading="CLI" group="cli">
+
+@include 'os/setup/cli.mdx'
```
fairclothjm, line 39: I disagree with the use of partials in this PR. It is only used here, so why put it in a partial? It makes it cumbersome to read this file because you have to open many files. It also makes it harder to search the docs.

drewmullen: > It is only used here, so why put it in a partial? 

Can you please specify what you mean by that? If its because partials are not used elsewhere in the docs - id say thats something we should remedy! its quite helpful

If you mean that the `vault_plugin` is only used once in deployment - id that the point of TFVP is to manage vault's configuration - many things are only used once

jonathanfrappier:
Why is this not being documented with the Vault provider docs? https://github.com/hashicorp/terraform-provider-vault/tree/main/website/docs/d - we'll end up with duplicate docs that need to both be updated in the event of a change

content/vault/v2.x/content/docs/secrets/os.mdx
fairclothjm: I agree that partials are helpful, but we shouldn't make a partial for something unless we have a good reason to. It spreads the information across multiple files. That makes it more burdensome to read and edit this file. It also makes it harder to grep the docs. In this PR, none of the partials are used more than once. Maybe we could [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)?

fairclothjm:
@jonathanfrappier Hello!

> Why is this not being documented with the Vault provider docs? https://github.com/hashicorp/terraform-provider-vault/tree/main/website/docs/d - 

We do have a docs PR open in TFVP. Personally, I support having a TF config example in these docs.

> we'll end up with duplicate docs that need to both be updated in the event of a change

If the API changes to this plugin, that would be true whether we have TFVP examples here or not. It is very rare to have TFVP API changes only.

schavis:
@drewmullen If we're not ready to merge, please mark the PR as a draft. We can still review it and convert it back when you're ready to release the update.

jonathanfrappier:
Close per discussion in Slack. Links to the TF Registry docs have been added to most of the auth methods and secrets engines https://developer.hashicorp.com/vault/docs/secrets/os#terraform

yhyakuna:
I'm re-opening this for further discussion.

yhyakuna:
@drewmullen I created https://github.com/hashicorp/web-unified-docs/pull/3463 to make some editorials to your original changes. Please take a look.

PR #2298, opened 2026-04-22
Upload the changed mdx files to artifacts to avoid arg limits

RubenSandwich:
## Description
Fixes the PR preview build link checker to support arbitrary long args. We do this by saving and uploading the changed MDX files as an artifact and then download and using that artifact file when running the link checker. Which means that the arg size will no longer be variable.

## Testing

This PR was tested in the PR: https://github.com/hashicorp/web-unified-docs/pull/2299.

### DIFF
--- a/.github/workflows/build-pr-preview.yml
+++ b/.github/workflows/build-pr-preview.yml
@@ -370,7 +370,7 @@ jobs:
 
       - name: Run lychee link checker for internal links
         id: lychee-internal
-        uses: lycheeverse/lychee-action@f613c4a64e50d792e0b31ec34bbcbba12263c6a6 # v2.3.0
+        uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411 # v2.8.0
         with:
           output: ./internal.md
           args: >-
@@ -394,7 +394,7 @@ jobs:
 
       - name: Run lychee link checker for external links
         id: lychee-external
-        uses: lycheeverse/lychee-action@f613c4a64e50d792e0b31ec34bbcbba12263c6a6 # v2.3.0
+        uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411 # v2.8.0
         with:
           output: ./external.md
           args: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/build-pr-preview.yml
```diff
@@ -350,18 +357,24 @@ jobs:
...
+          cat ${{ github.workspace }}/tmp/changed-mdx-files.txt
 
       - name: Run lychee link checker for internal links
         id: lychee-internal
         uses: lycheeverse/lychee-action@f613c4a64e50d792e0b31ec34bbcbba12263c6a6 # v2.3.0
         with:
           output: ./internal.md
           args: >-
-            ${{ needs.get-changed-files.outputs.changed_markdown_content_files }}
+            --files-from ${{ github.workspace }}/tmp/changed-mdx-files.txt
```
williamdalessandro, line 377: Hey was going through the test output for this PR and noticed that there's an [error with the `--files-from` argument here](https://github.com/hashicorp/web-unified-docs/actions/runs/24800935359/job/72584174070?pr=2299#step:5:93) so it doesn't seem like lychee is actually running on the files for both internal/external links- would you be able to take a closer look into that?

williamdalessandro (changes requested):
Great job on this @RubenSandwich ! I like the approach but noticed an error with the argument used to call lychee https://github.com/hashicorp/web-unified-docs/actions/runs/24800935359/job/72584174070?pr=2299#step:5:93

.github/workflows/build-pr-preview.yml
```diff
@@ -103,7 +110,7 @@ jobs:
   deploy-unified-docs-api-preview:
     name: Deploy Unified Docs API Preview
     runs-on: ubuntu-latest
-    needs: [security-check]
+    needs: [security-check, get-changed-files]
```
RubenSandwich, line 113: This isn't required for this PR, but both `security-check` and `get-changed-files` should succeeded before moving forward. I think we just make a mistake before and forgot this. So I'm gonna try to sneak it into this PR.

williamdalessandro (approved):
The changes look good and the testing branch output looks solid as well- nice job!

PR #2312, opened 2026-04-25
Consul: drop bootstrap and ui_config from client multi-interface example

alliasgher:
## Summary

The Consul docs page `Agents → Configuration` includes an example titled "Client node with multiple interfaces or IP addresses" whose body sets `server = false` (correct, it's a client) but also sets `bootstrap = true` and `ui_config.enabled = true`.

`bootstrap` is server-only: it tells the single bootstrap server to elect itself leader, so it makes no sense on a client agent. The `ui_config` block is also unrelated to what the example is meant to demonstrate (the multi-interface `bind_addr`/`client_addr` setup) and looks like server-side config on a section that is supposed to be about clients. That mismatch is exactly the confusion reported in [hashicorp/consul#18523](https://github.com/hashicorp/consul/issues/18523).

This PR drops both `bootstrap` and `ui_config` from:

- the canonical `global/partials/examples/agent/client/multi-interface.mdx` partial used by current versions, and
- the inlined copy in `v1.20.x/content/docs/agent/index.mdx` that doesn't go through the partial.

No other inputs are changed; the example still demonstrates the same multi-interface configuration.

### DIFF
--- a/content/consul/global/partials/examples/agent/client/multi-interface.mdx
+++ b/content/consul/global/partials/examples/agent/client/multi-interface.mdx
@@ -11,12 +11,8 @@ information](/consul/docs/reference/agent/configuration-file/general#client_addr
 <CodeTabs>
 
 ```hcl
-node_name = "consul-client"
-server    = false
-bootstrap = true
-ui_config {
-  enabled = true
-}
+node_name  = "consul-client"
+server     = false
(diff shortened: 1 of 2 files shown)


aimeeu (approved):
Thanks for fixing this!

PR #2331, opened 2026-04-28
HCP Boundary Content Relocation

rmainwork:
Re-run of #1902. An integration branch was created to keep all the HCP work together (and to make integration easier). In the process, the original (confusing) branch name was changed which caused the PR to be closed (by GitHub)...so this PR was (re)created with the same commits for that reason.

PR #2332, opened 2026-04-28
HCP vault radar backend

rmainwork:
Break HCP Vault Radar out as it's own product (rather than inside HCP Docs) as-per the RFC .

- [RFC](https://docs.google.com/document/d/1VPsg3TfFZG129h2ZoBZxQvn_bWJW4bL5dIzrDW0iqug/edit?tab=t.0)
- [Monday.com task](https://ibm.monday.com/boards/18401975210/pulses/10908213435)

### Testing
- Hit https://web-unified-docs-99m60qind-hashicorp.vercel.app/hcp/docs/vault-radar/security-model. You should get docs back for Vault Radar security model


RubenSandwich (approved):
Metadata looks good. 👍🏻

yhyakuna:
Shouldn't the folder be `hcp-vault-radar` instead of `vault-radar-hcp` to be consistent? (HCP Boundary is `content/hcp-boundary` in [PR 2331](https://github.com/hashicorp/web-unified-docs/pull/2331/changes#diff-a22c263686553013feaeb0677d26eeb0b8778a756c4311c1fce13384258026aa).)

rmainwork:
@yhyakuna ah - Good point! I went with `vault-radar-hcp` to match the RFC.....but #2331 may need to be changed to be consistent with that.... What do you think?

yhyakuna:
> @yhyakuna ah - Good point! I went with `vault-radar-hcp` to match the RFC.....but #2331 may need to be changed to be consistent with that.... What do you think?

I just double-checked with the tech writers, and yes, the folder name should be `<product>-hcp`, and should be consistent all across. So, I guess https://github.com/hashicorp/web-unified-docs/pull/2331 needs to be updated.

rmainwork:
@RubenSandwich when you have a sec - would you mind giving 5ede0b822a5c1f44c83e702e11b9045749ab6d6e the once over?

RubenSandwich (changes requested):
Vault radar seems to be broken: https://unified-docs-frontend-preview-knqg6xc43-hashicorp.vercel.app/hcp/docs/vault-radar.

[I do not think you need to define a basePath for](https://github.com/hashicorp/web-unified-docs/pull/2332/changes#diff-11cda08053ebc99044ec6fd57d21709c6f846439350623471c9e54986faed281R477) `vault-radar-hcp` as the docs are in the root version folder.

jonathanfrappier:
Is this effort proceeding? I was asked to populate the landing pages but want to do it at the correct time to avoid re-work if possible.

rmainwork:
@jonathanfrappier I believe so - but we're currently working on a datadog/sentry migration before that gets sunset. Plus I believe we might be waiting for some feedback (from the education team? not 100% sure though) - @emilypersson1  might have more context on where things are with this.

emilypersson1:
@jonathanfrappier like Robert said, we've been working on a migration that's pretty time intensive and critical priority, so we haven't had a chance to work on this too much. we're reassigning this to someone to pick up as soon as there's capacity, so there should be more movement on this in the next week or two. apologies for the delay on our end! i'll follow up once we're ready to resume work here

PR #2342, opened 2026-04-29
reapply 1.14 nav and post-apply changes

trujillo-adam:
This PR reapplies the nav changes from v1.14.x that were apparently made after the v1.15.x folder was copied.
It also re-adds the "Perform post-apply operations" topic, which had been mixed into the topic on using provisioners but was spun off in the v1.14.x version of the docs.

### DIFF
--- a/content/terraform/v1.15.x/data/language-nav-data.json
+++ b/content/terraform/v1.15.x/data/language-nav-data.json
@@ -1,73 +1,12 @@
 [
   {
-    "heading": "Terraform Language"
+    "title": "Get started",
+    "href": "https://developer.hashicorp.com/terraform/tutorials"
   },
   {
-    "title": "Overview",
-    "path": ""
-  },
-  {
-    "title": "Attributes as Blocks - Configuration Language",
-    "path": "attr-as-blocks",
-    "hidden": true
-  },
-  {
-    "title": "Style guide",
-    "path": "style"
-  },
-  {
-    "title": "Syntax",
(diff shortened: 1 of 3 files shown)


content/terraform/v1.15.x/data/language-nav-data.json
```diff
@@ -277,35 +208,76 @@
...
+      { "title": "Use cases", "path": "stacks/use-cases" },
+      { "title": "Design a Stack", "path": "stacks/design" },
       {
-        "title": "Validate your configuration",
-        "path": "validate"
+        "title": "Create a Stack",
+        "routes": [
+          { "title": "Define configuration", "path": "stacks/component/config" },
+          { "title": "Declare providers", "path": "stacks/component/declare-providers" },
+          { "title": "Manage components", "path": "stacks/component/manage" }
```
mattschouten: Noting this is on line 273—at the same nav level as "Create a Stack" or "Define deployments"—in 1.14.

Which way did we like better?

mattschouten (approved):
v1.14 and v1.15 `post-apply-operations.mdx` and `provisioners.mdx` are identical ✅

Found one difference on the language nav, and don't remember which way we wanted it to be.

content/terraform/v1.15.x/data/language-nav-data.json
trujillo-adam: Ah, yeah, it should be at the same level, good catch

PR #2359, opened 2026-05-01
Allow setting dev-portal ref overrides

RubenSandwich:
### Description

Allow overrides of dev portal build through a special PR body string. Due to security reasons this ref needs to exist in dev-portal-internal.

### Review & Testing

1. Double check that the pull request target build used the correct repo and ref: https://github.com/hashicorp/web-unified-docs/actions/runs/25226639306/job/73972001317#step:3:69.
2. Double check that the dev-portal preview homepage matches the content in the dev-portal-internal ref: https://github.com/hashicorp/dev-portal-internal/compare/main...rn/test-UDR-target.
3. Double check for any security concerns I might of missed.

### Dev portal overrides

/set-env USE_DEV_PORTAL_INTERNAL_REF=rn/test-UDR-target


williamdalessandro (approved):
Hey @RubenSandwich looked through the testing steps and it all looked good! Was thinking over the security concerns and couldnt think of much behind people from forked repos potentially guessing branch names in UDR internal to see early content, but I honestly don't think that's worth much to stop me from approving this

PR #2407, opened 2026-05-11
nomad: new `max_run_duration` field and associated metrics documentation

pkazmierczak:
## Description

Documentation for a new feature of (sys)batch jobs: allowing users to set their maximum duration. This lands in 2.0.2. 

## Links

Nomad Code PR: https://github.com/hashicorp/nomad/pull/27803
Jira: https://hashicorp.atlassian.net/browse/NMD-551
GitHub Issue: https://github.com/hashicorp/nomad/issues/1782

## Contributor checklists

Review urgency:

- [ ] ASAP: Bug fixes, broken content, imminent releases
- [x] 3 days: Small changes, easy reviews
- [ ] 1 week: Default expectation
- [ ] Best effort: No urgency

Pull request:

- [x] Verify that the PR is set to merge into the correct base branch
- [ ] Verify that all status checks passed
- [ ] Verify that the preview environment deployed successfully
- [ ] Add additional reviewers if they are not part of assigned groups

Content:

- [x] I added redirects for any moved or removed pages
- [x] I followed the [Education style guide](https://github.com/hashicorp/web-unified-docs/tree/main/docs/style-guide)
- [x] I looked at the local or Vercel build to make sure the content rendered correctly

## Reviewer checklist

- [ ] This PR is set to merge into the correct base branch.
- [ ] The content does not contain technical inaccuracies.
- [ ] The content follows the Education content and style guides.
- [ ] I have verified and tested changes to instructions for end users.

[NMD-1234]: https://hashicorp.atlassian.net/browse/NMD-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/nomad/v2.0.x/content/docs/job-specification/group.mdx
+++ b/content/nomad/v2.0.x/content/docs/job-specification/group.mdx
@@ -58,16 +58,15 @@ job "docs" {
   regains connectivity is also specified here.
 
 - `max_run_duration` `(string: "")` - Specifies the maximum duration for which
-  allocations in this group are allowed to run. Only valid for `batch` and
+  each allocation in this group is allowed to run. Only valid for `batch` and
   `sysbatch` jobs. When the duration is exceeded, all tasks in the group are
-  killed and the allocation is marked as `complete` with a client description
-  of `allocation exceeded max_run_duration`. The job status transitions to
-  `dead`.
+  killed and the allocation is marked as `complete` with a client description of
+  `allocation exceeded max_run_duration`. 
 
-  The timer begins immediately when the allocation is created in the allocation
-  runner, regardless of individual task states. Tasks that take longer to start
-  than the configured timeout will therefore also be terminated — including
-  allocations that never finish downloading images or artifacts.
+  The timer begins immediately when the allocation is created in the client,
+  regardless of individual task states. Tasks that take longer to start than the
+  configured timeout will therefore also be terminated — including allocations
+  that never finish downloading images or artifacts.
 
   `prestart` and `poststart` lifecycle task runtimes count toward
   `max_run_duration`. `poststop` tasks will **not** be started if the
--- a/content/nomad/v2.0.x/content/docs/job-specification/group.mdx
+++ b/content/nomad/v2.0.x/content/docs/job-specification/group.mdx
@@ -73,6 +73,20 @@ job "docs" {
   `max_run_duration`. `poststop` tasks will **not** be started if the
   allocation is stopped due to this timeout.
 
+  Task [restarts][restart] do not reset the timer — once a deadline is
+  established it is not recalculated unless `max_run_duration` itself is
+  changed in a job update. All restart attempts therefore consume time from
+  the same budget. If the Nomad client process restarts mid-countdown, the
+  original deadline is reconstructed from each task runner's persisted
+  `StartedAt` timestamp, so the remaining budget is honoured rather than
+  reset to the full duration.
+
+  When the timeout fires, the allocation is marked `complete` rather than
+  `failed`, so the group's [`reschedule`][reschedule] policy is not
+  triggered. If an allocation is rescheduled for an independent reason (for
+  example, a node failure before the timeout), the replacement allocation
+  starts its own `max_run_duration` timer from zero.
+
 - `meta` <code>([Meta][]: nil)</code> - Specifies a key-value map that annotates
   with user-defined metadata.
 
--- a/content/nomad/v2.0.x/content/docs/job-specification/group.mdx
+++ b/content/nomad/v2.0.x/content/docs/job-specification/group.mdx
@@ -57,6 +57,22 @@ job "docs" {
   when the client disconnects. The policy for reconciliation in case the client
   regains connectivity is also specified here.
 
+- `max_run_duration` `(string: "")` - Specifies the maximum duration for which
+  allocations in this group are allowed to run. Only valid for `batch` and
+  `sysbatch` jobs. When the duration is exceeded, all tasks in the group are
+  killed and the allocation is marked as `complete` with a client description
+  of `allocation exceeded max_run_duration`. The job status transitions to
+  `dead`.
+
+  The timer begins immediately when the allocation is created in the allocation
+  runner, regardless of individual task states. Tasks that take longer to start
+  than the configured timeout will therefore also be terminated — including
+  allocations that never finish downloading images or artifacts.
+
+  `prestart` and `poststart` lifecycle task runtimes count toward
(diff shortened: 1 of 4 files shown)


content/nomad/v2.0.x/content/docs/job-specification/group.mdx
```diff
@@ -57,6 +57,22 @@ job "docs" {
...
+  allocations in this group are allowed to run. Only valid for `batch` and
+  `sysbatch` jobs. When the duration is exceeded, all tasks in the group are
+  killed and the allocation is marked as `complete` with a client description
+  of `allocation exceeded max_run_duration`. The job status transitions to
+  `dead`.
+
+  The timer begins immediately when the allocation is created in the allocation
+  runner, regardless of individual task states. Tasks that take longer to start
+  than the configured timeout will therefore also be terminated — including
+  allocations that never finish downloading images or artifacts.
```
tgross: I'd also include here an explanation around how the timer works with client restarts or down clients, task restarts, and allocation reschedules (it maybe seems obvious to you or me how this works with reschedules, but restart vs reschedule behavior is a frequent source of confusion)

aimeeu (approved):
Approved but deferring to Tim on technical.

tgross (approved):
LGTM assuming we land https://github.com/hashicorp/nomad/pull/27803 for 2.0.1

aimeeu:
I added a release note to this PR just in case the code PR merges for the 2.0.1 release.

content/nomad/v2.0.x/content/docs/job-specification/group.mdx
```diff
@@ -57,6 +57,36 @@ job "docs" {
   when the client disconnects. The policy for reconciliation in case the client
   regains connectivity is also specified here.
 
+- `max_run_duration` `(string: "")` - Specifies the maximum duration for which
+  allocations in this group are allowed to run. Only valid for `batch` and
```
schmichael: Instead of...

> Specifies the maximum duration for which allocations in this group are allowed to run.

...maybe:

> Specifies the maximum duration for which *each allocation* in this group *is* allowed to run.

What I'm trying to clarify is that the deadline applies to each allocation individually. This is particularly crucial if not all allocations can be placed at once and some are started later than others. In that case each allocation will get their own deadline based on this duration (and **not** calculate their deadline based off of the first placement made or job creation or something else).

There's so many different ways this feature could work, and I just know one of us is going to be squinting as these docs 4 years from now trying to figure out what we intended some subtle behavior to be. 😅

content/nomad/v2.0.x/content/docs/job-specification/group.mdx
```diff
@@ -57,6 +57,36 @@ job "docs" {
...
+  `sysbatch` jobs. When the duration is exceeded, all tasks in the group are
+  killed and the allocation is marked as `complete` with a client description
+  of `allocation exceeded max_run_duration`. The job status transitions to
+  `dead`.
```
schmichael:
```suggestion
  of `allocation exceeded max_run_duration`.
```

Does `max_run_duration` impact job status directly or only as a side effect of allocations becoming terminal? If it's just the normal "job status is `dead` when all allocations are `complete`", then let's not mention it here. I don't want users to think setting a deadline changes the (already confusing!) job status logic.

If I'm missing something and the deadline transitioning allocs to `complete` does transition jobs to `dead` when they normally wouldn't: then it's probably worth an extra phrase here describing how this is a new flow.

content/nomad/v2.0.x/content/docs/job-specification/group.mdx
```diff
@@ -57,6 +57,36 @@ job "docs" {
...
+
+  The timer begins immediately when the allocation is created in the allocation
+  runner, regardless of individual task states. Tasks that take longer to start
```
schmichael: > The timer begins immediately when the allocation is created in the ~~allocation runner,~~ _client before the tasks are started._

I don't think we have ever used "allocation runner" in docs! I think "client" will more immediately communicate to users what we intend. While task states are exposed to users, I think the "happens-before" aspect is the most critical thing to call out.

content/nomad/v2.0.x/content/docs/job-specification/group.mdx
```diff
@@ -57,6 +57,36 @@ job "docs" {
...
+  `max_run_duration`. `poststop` tasks will **not** be started if the
+  allocation is stopped due to this timeout.
+
+  Task [restarts][restart] do not reset the timer — once a deadline is
+  established it is not recalculated unless `max_run_duration` itself is
+  changed in a job update. All restart attempts therefore consume time from
+  the same budget. If the Nomad client process restarts mid-countdown, the
+  original deadline is reconstructed from each task runner's persisted
+  `StartedAt` timestamp, so the remaining budget is honoured rather than
+  reset to the full duration.
```
schmichael, line 81: Oof I always forget the client restart case and it changes everything because you suddenly need deterministic logic based on durable state!

What if we switched the logic to always calculate the deadline based on `Allocation.CreateTime + TaskGroup.MaxRunDuration`?

**Pros**
1. Easy for users to calculate in their head just looking at the job + alloc
2. Stable across client restarts
3. No new state
4. Matches the timespan the server observes the allocation's resources as used (so using a gpu with max_run_duration=1h means an allocation gets a gpu for 1 hour _from the scheduler's perspective._ That's arguably the safest way to model it.)

**Cons**
1. Clock skew between servers and clients now impacts workload lifetimes which is a new and unfortunate behavior. We really try to keep clock logic local other places for this reason.
2. What's the latency between a server creating an allocation and a client running it? How does a jobspec author know that number to take it into account when choosing a `max_run_duration`?

Operators should strive to minimize both Cons 1 & 2 (sub-second for both!), so I think the pros outweigh the cons here. Users should _not_ expect even second-level granularity from this feature regardless of what we choose as the start time.

(Left an internal comment too.)

schmichael (commented):
Looks good so far but sounds like we need to work out client restart logic before proceeding.

pkazmierczak:
I'll address Schmichael's comments but just a heads up @aimeeu this won't be landing on 2.0.1.

content/nomad/v2.0.x/content/docs/job-specification/group.mdx
pkazmierczak: Ok. Yeah. I think this is the way I should've done this from the start. See https://github.com/hashicorp/nomad/pull/27803/changes/5ac3e359f0fef36935dfb6f52559b3f42d1beb11 for implementation (it actually simplifies things a lot).

PR #2416, opened 2026-05-12
[VAULT] Add missing redirect and tweak older ones

schavis:
Add redirect for old LTS support overview and tweaks older redirect to point at the new support page


mister-ken (approved):
lgtm

PR #2423, opened 2026-05-12
Document the manual stack configuration api flow

aaabdelgany:
### Terraform


content/terraform-docs-common/docs/cloud-docs/api-docs/stacks/configurations.mdx
```diff
@@ -573,17 +579,82 @@ $ curl\
...
+
+```shell
+$ curl\
+  --header "Content-Type: application/octet-stream" \
+  --request PUT \
+  --data-binary @configuration.tar.gz \
+  "$UPLOAD_URL"
+```
+
+Replace `$UPLOAD_URL` with the value of `upload-url` from the [Fetch a Stack configuration upload URL](#fetch-stack-configuration-upload-url) response.
```
jondavidjohn: I think this should be `source-upload-url` right?

content/terraform-docs-common/docs/cloud-docs/api-docs/stacks/configurations.mdx
```diff
@@ -573,17 +579,82 @@ $ curl\
...
+
+The response includes a `source-upload-url` field containing a temporary URL for uploading the configuration archive.
+
 ```json
 {
   "data": {
-    "source-upload-url": "https://github.com/stack-config-uploads"
+    "source-upload-url": "https://archivist.terraform.io/v1/object/..."
   },
   "links": {
```
jondavidjohn: How super sure are we that the `links` get added? 🤔

aaabdelgany: they don't, nice catch!

jondavidjohn (approved):

trujillo-adam:
Hey @aaabdelgany , when do we need to publish this? Is this the only way to accomplish this task? Would other users _not_ building integrations with HCP Terraform also need or want to do this?  

I have some concerns that we're mixing the content types a little too much. In order to continue implementing the [Diataxis](https://diataxis.fr) framework, I think we should move the "Complete manual upload workflow" section out of the reference docs and into the usage docs. We can cross-link the topics to connect the them. This is 1) to help users who aren't going directly to the API reference find this information (if it's important to their success) and 2) help keep different types of information indexed separately to improve maintainability. 

There isn't a really great place to move this topic to though. From a user journey perspective, I think I would put it in the "Create a Stack" section in the language docs just after the [Define a configuration](https://developer.hashicorp.com/terraform/language/stacks/component/config). That seems to be the most logical place a user would want to encounter this topic as they are exploring the instructions. But because the language docs are versioned and not really about HCP TF, things get a little messy.

The next best place seems to be in the "Manage a Stack" section in the HCP TF docs probably after [Configure a Stack](https://developer.hashicorp.com/terraform/cloud-docs/stacks/configure).

Does that make sense?

aaabdelgany:
> Hey @aaabdelgany , when do we need to publish this? Is this the only way to accomplish this task? Would other users _not_ building integrations with HCP Terraform also need or want to do this?
> 
> I have some concerns that we're mixing the content types a little too much. In order to continue implementing the [Diataxis](https://diataxis.fr) framework, I think we should move the "Complete manual upload workflow" section out of the reference docs and into the usage docs. We can cross-link the topics to connect the them. This is 1) to help users who aren't going directly to the API reference find this information (if it's important to their success) and 2) help keep different types of information indexed separately to improve maintainability.
> 
> There isn't a really great place to move this topic to though. From a user journey perspective, I think I would put it in the "Create a Stack" section in the language docs just after the [Define a configuration](https://developer.hashicorp.com/terraform/language/stacks/component/config). That seems to be the most logical place a user would want to encounter this topic as they are exploring the instructions. But because the language docs are versioned and not really about HCP TF, things get a little messy.
> 
> The next best place seems to be in the "Manage a Stack" section in the HCP TF docs probably after [Configure a Stack](https://developer.hashicorp.com/terraform/cloud-docs/stacks/configure).
> 
> Does that make sense?

👋 @trujillo-adam my apologies for the delay. I agree that the `complete manual upload workflow` section feels more like a tutorial-like reference as opposed to something appropriate for api docs. 

Moving it to the `create a stack` section in the usage docs might be the most logical place atm but I don't want users to get the impression that manual stack configuration is the primary workflow for config creation. It might be better to eventually make some sort of master page walking through the various methods - but I'm kind of leaning towards removing the `complete manual upload workflow` section entirely since that isn't something we have right now

trujillo-adam (approved):
Looks good.

PR #2429, opened 2026-05-13
ci(force-rebuild-dev-portal): deny-all permissions

arpitjain099:
`force-rebuild-dev-portal.yml` has no checkout step and makes no GitHub API calls — it just runs `curl -X POST ${{ secrets.DEV_PORTAL_DEPLOY_HOOK_PROD }}`. `permissions: {}` (deny-all) is the most accurate scope for the default `GITHUB_TOKEN` here.

YAML validated with `yaml.safe_load`.


arpitjain099:
Closing this for now. Thanks for the review consideration; may revisit in the future if needed.

PR #2448, opened 2026-05-14
feat: use merge-base instead of base sha for inc-builds

LeahMarieBush:
### Description

This PR changes the `mergeBase` for incremental builds. Previously, it used the `BASE_SHA` for a PR to determine which files were changed. However, this does not take into account changes from the branch it is based off of if that branch is not main. Now it is changed to the merge-base of the PR to account for all changes.

[Asana task](https://ibm.monday.com/boards/18401913557/pulses/11995631626)

### Testing

I have a [test PR](https://github.com/hashicorp/web-unified-docs/pull/2449) that shows the changed files for both this PR and that one. There are testing instructions in that PR description you can follow.

### Note

The build preview will fail for this PR since the `mergeBase` requires a fetch-depth of 0 but the build preview action does not yet have this (since it used pull_request_target). You can see that it will succeed in the testing PR


RubenSandwich (approved):
🚀

PR #2486, opened 2026-05-21
Changed HCP terraform  to Terraform Enterprise

munnep:
# Terraform Enterprise 2.x.x

The health page was consistently mentioning HCP Terraform instead of Terraform Enterprise. This caused confusion. Altered the documentation page to mention Terraform Enterprise. 

For HCP terraform there is another page.


munnep:
Needed to do it on another repo

PR #2490, opened 2026-05-21
VAULT-42829: Create a new billing config endpoint to set custom retention months

aslamovamir:
This PR adds updates to the docs about a new billing config endpoint we added to let customers set custom retention months for their consumption billing data.

Jira: https://hashicorp.atlassian.net/browse/VAULT-42829

Vault PR: https://github.com/hashicorp/vault-enterprise/pull/14785

Screenshot 2026-05-21 at 3 26 57 PM
Screenshot 2026-05-21 at 3 31 38 PM
Screenshot 2026-05-21 at 3 32 30 PM
Screenshot 2026-05-21 at 3 32 45 PM
Screenshot 2026-05-21 at 3 32 50 PM

PR #2494, opened 2026-05-22
VAULT-42829: Create a new billing config endpoint to set custom retention months

aslamovamir:
This PR adds updates to the docs about a new billing config endpoint we added to let customers set custom retention months for their consumption billing data.

Jira: https://hashicorp.atlassian.net/browse/VAULT-42829

Vault PR: https://github.com/hashicorp/vault-enterprise/pull/14785
Original Docs PR (closed in favor of this due to base change effects): https://github.com/hashicorp/web-unified-docs/pull/2490

Screenshot 2026-05-21 at 3 26 57 PM
Screenshot 2026-05-21 at 3 31 38 PM
Screenshot 2026-05-21 at 3 32 30 PM
Screenshot 2026-05-21 at 3 32 45 PM
Screenshot 2026-05-21 at 3 32 50 PM

### DIFF
--- a/content/vault/v2.x/content/api-docs/system/billing/config.mdx
+++ b/content/vault/v2.x/content/api-docs/system/billing/config.mdx
@@ -17,7 +17,11 @@ The `/sys/billing/config` endpoint allows you to configure how long Vault retain
 
 By default, Vault retains `sys/billing/overview` data for 37 months, including the current month. You can configure the retention period to be between 13 months (minimum) and 72 months (maximum).
 
-Increasing the retention period cannot restore historical data for months that was previously outside the retention period, but will retain data from the previous retention period.
+Vault does not restore historical data when you increase the retention period.
+Instead, Vault accumulates new data up to the new retention period before
+advancing the retention window. For example, if you increase your retention
+window from 13 months to 37 months, Vault does not advance the retention window
+until it accumulates 37 months of data.
 
 
 ## Read billing configuration
--- a/content/vault/v2.x/content/api-docs/system/billing/config.mdx
+++ b/content/vault/v2.x/content/api-docs/system/billing/config.mdx
@@ -15,7 +15,7 @@ last_modified: 2026-05-21T14:00:00-05:00
 
 The `/sys/billing/config` endpoint allows you to configure how long Vault retains `/sys/billing/overview` data.
 
-By default, Vault retains billing data for 37 months, including the current month. You can configure the retention period to be between 13 months (minimum) and 72 months (maximum).
+By default, Vault retains `sys/billing/overview` data for 37 months, including the current month. You can configure the retention period to be between 13 months (minimum) and 72 months (maximum).
 
 Increasing the retention period cannot restore historical data for months that was previously outside the retention period, but will retain data from the previous retention period.
 
--- a/content/vault/v2.x/content/api-docs/system/billing/config.mdx
+++ b/content/vault/v2.x/content/api-docs/system/billing/config.mdx
@@ -13,7 +13,7 @@ last_modified: 2026-05-21T14:00:00-05:00
 
 @include 'alerts/restricted-root.mdx'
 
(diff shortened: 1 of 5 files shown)


content/vault/v2.x/content/api-docs/system/billing/config.mdx
```diff
@@ -0,0 +1,89 @@
...
+created_at: 2026-05-21T14:00:00-05:00
+last_modified: 2026-05-21T14:00:00-05:00
+# END AUTO GENERATED METADATA
+---
+
+# `/sys/billing/config`
+
+@include 'alerts/restricted-root.mdx'
+
+The `/sys/billing/config` endpoint allows you to configure how long Vault retains billing data.
```
gsantos-hc: I'm assuming the retention period only applies to `sys/billing/overview` data? Or does it extend to certificate counts (`sys/billing/certificates`) and client counts?


```suggestion
The `/sys/billing/config` endpoint allows you to configure how long Vault retains `/sys/billing/overview` data.
```

content/vault/v2.x/content/api-docs/system/billing/config.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+By default, Vault retains billing data for 37 months, including the current month. You can configure the retention period to be between 13 months (minimum) and 72 months (maximum).
```
gsantos-hc:
```suggestion
By default, Vault retains `sys/billing/overview` data for 37 months, including the current month. You can configure the retention period to be between 13 months (minimum) and 72 months (maximum).
```

content/vault/v2.x/content/api-docs/system/billing/config.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+Increasing the retention period cannot restore historical data for months that was previously outside the retention period, but will retain data from the previous retention period.
```
gsantos-hc: Don't love this phrasing. @schavis Do you have any suggestions?

We're trying to say here that:
- If the customer was retaining data for 13 months, changing this to 37 months won't instantly make 37 months of data show up.
- As time goes by, after the user increases retention period, Vault will accumulate up to 37 months of data.

gsantos-hc (commented):
Couple of suggested edits.

Also: Are we releasing this in the next patch version, or will it go into 2.1 (October)? If 2.1, we need to update the target branch.

schavis (approved):

content/vault/v2.x/content/api-docs/system/billing/config.mdx
schavis:
```suggestion
Vault does not restore historical data when you increase the retention period.
Instead, Vault accumulates new data up to the new retention period before
advancing the retention window. For example, if you increase your retention
window from 13 months to 37 months, Vault does not advance the retention window
until it accumulates 37 months of data.
```

schavis: I added a suggestion below. Let me know what you think

aslamovamir:
> Couple of suggested edits.
> 
> Also: Are we releasing this in the next patch version, or will it go into 2.1 (October)? If 2.1, we need to update the target branch.

Not sure exactly which version it should go to, my thinking is 2.0.2 is fine? so just miner release, instead of patch? The vault side 2.0.2 will include this endpoint

content/vault/v2.x/content/api-docs/system/billing/config.mdx
aslamovamir: Good point, yes! it only applies to sys/billing/overview endpoint

content/vault/v2.x/content/api-docs/system/billing/config.mdx
aslamovamir: I like it, @gsantos-hc if you like this too, we can commit her suggestion:)

gsantos-hc (approved):

schavis (approved):

PR #2498, opened 2026-05-22
[VAULT] Add a deploy to docker guide

schavis:
Extract the Docker instructions from the ZCX guide to have a Docker-specific deploy guide


content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+- Your encryption certificates, under `vault-deploy/local-config/certs`.
+- Your Vault node configuration files under `vault-deploy/local-config/hcl`.
+- Your Docker compose and `Dockerfile` files for the load balancer under
+  `proxy-deploy`.
+- Your load balancer configuration files under `proxy-deploy/local-config`
+
+
+## Step 3: Create a Docker environment file for Vault
+
+To simplify maintenance, we recommend creating an single file,
```
yhyakuna:
```suggestion
To simplify maintenance, we recommend creating a single file,
```

yhyakuna (approved):
I found a minor typo, but other than that, it looks good to me.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+
+RUN mkdir /vault/data/
+RUN mkdir /vault/plugins/
+RUN mkdir /vault/plugins/tmp
+
+# Make sure Vault owns the Vault-related files
+RUN chown -R vault:vault /vault
+
+# Set ownership and mode for certificate files
+RUN chown root:vault /vault/certs/vault-key.pem
```
brianshumate: I'm afraid this is inaccurate. The `vault.Dockerfile` COPY commands place the TLS material in the path `/vault/config/certs/`, but these commands operate against the path `/vault/certs/`.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+          NODE_IDX:   "3"
+
+      include:
+        - path: ../shared-deploy/cluster-env.yml
+    ```
+
+
+## Step 8: Create your Vault configuration files
+
+Use the following template to create individual configuration files
```
brianshumate, line 304: It could be helpful to mention here that these configuration files must be created within the `local-config/hcl/` path.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+# syntax=docker/dockerfile:1
+FROM hashicorp/vault-enterprise:<version_tag>
+
+# Copy the config and license files
+COPY local-config/hcl/vault-*.hcl     /vault/config/hcl/
+COPY local-config/vault-license.hclic /vault/config/vault-license.hclic
+
+# Copy the cert files
+COPY local-config/certs/vault.pem /vault/config/certs/vault.pem
+COPY local-config/certs/vault.key /vault/config/certs/vault.keyt
```
brianshumate: There's a typo in the target path (`vault.keyt`). I would suggest the fix, but do not have the ability to do so for this PR.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+# --- Integrated storage ---
+storage "raft" {
+
+  path    = "/vault/data"
+  node_id = "<container_name>"
+
+  # Rejoin/listener configuration for cluster node N+1
+  retry_join {
+    leader_api_addr    = "https://<alt_container_name_1>:8200"
+    tls_disable        = "false"
```
brianshumate: Unless I'm mistaken or the API docs are outdated, the [retry_join](https://developer.hashicorp.com/vault/docs/configuration/storage/raft#retry_join-stanza) stanza does not accept `tls_` configuration parameters, but uses a different `leader_` syntax to specify the TLS material values.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+directory information. Even if you currently do not intend to register custom or
+enterprise plugins, we recommend deciding on a plugin location at setup because
+you must restart the cluster to update the plugin information later.
+
+For example:
+
+```hcl
+ui = true
+disable_mlock = true
+license_path = "/vault/config/license.hclic"
```
brianshumate: Around line 215, the `vault.Dockerfile` copies the license file to `/vault/config/vault-license.hclic`, and I think this `license_path` value should match that value for the configuration to be valid.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+      vault-3:
+        extends:
+          service: vault-base
+        container_name: prod-vault-3
+        hostname: prod-vault-3
+        environment:
+          VAULT_ADDR: https://prod-vault-3:8200
+          NODE_IDX:   "3"
+
+      include:
```
brianshumate: It appears that in this version of the file, the `include:` block is indented too far, and should be on the same level as the `name:` and `services:` blocks to be considered valid YAML.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+
+## Step 5: Create a basic compose file for Vault
+
+Create a basic Docker compose file, `vault-deploy/vault.compose.yml` that
+defines the base build process for your Vault nodes:
+
+```yaml
+name: <project_name>
+services:
+  <serice_name>:
```
brianshumate: Typo in `<service_name>:`

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+    between nodes
+
+- **Optional: install the [Vault CLI](/vault/install/)**. To make Vault CLI
+  calls, you must install Vault locally. If you prefer not to install Vault, you
+  can make API calls from the command line instead.
+
+
+## Step 1: Create your persistent Docker volumes
+
+We recommend creating a persistent Docker volume for Vault and HAProxy
```
brianshumate: Not sure where the HAProxy is coming from, what it relates to, or why it appears here.

I think it's possibly related to the incomplete `proxy-deploy` piece? I might be worth removing all of that stuff and just generically mentioning that it's typical to front a Vault cluster with different types of proxies with layer 4 and layer 7 being the most common and leave it at that?

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
```diff
@@ -0,0 +1,589 @@
...
+    ```
+
+1. Use the initialization details to run `vault operator unseal` against the
+   container with the generated unseal keys. For example, to unseal the Vault
+   instance running in the `prod-vault-1` container:
+
+    ```shell-session
+    $ for token in $(
+        cat /secure/location/vault-init.json | \
+        jq -r '.unseal_keys_b64[0:3]         | join(" ")'
```
brianshumate: The use of `join(" ")` here is dubious. The `join(1)` shell command does not work this way, and is instead designed to perform SQL type joins on files (see also `man join`). I'm afraid this entire command line does not represent a legitimate way to unseal Vault in the manner described.

If `join(" ")` is omitted from the pipeline, the updated command does produce an output containing 3 of the unseal keys, but it needs to choose one of these to use for `vault operator unseal` (and a different one for each subsequent unseal operation).

```shell
cat vault-init.json | jq -r '.unseal_keys_b64[0:3]'
[
  "t9mPamRz/Gzhoy+............/mJjjzc/8h0yaR15a",
  "/Z0u+Or4SOqMe............iuCVmz+N1xvOSZ4tddhs",
  "3OgSYqF............g8p0ZSmPwT2tTW9WU/bT0c3w6"
]
```

brianshumate (commented):
While I didn't test this end to end, I came pretty close to that point, and I found some inconsistencies worth looking into.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
schavis: I just missed removing it from the zCX version of the doc

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
schavis: Supposedly someone from IBM ran through these steps end-to-end and it worked... I am starting to doubt that claim 🙃

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
schavis: moved

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
schavis: Added 👍

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
schavis: Fixed.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
schavis: I think this might be a difference in bash on Ubuntu vs MacOS. I know for a fact the command as written works from my previous testing and the functions I use on my own machine. 

The `join` here isn't the shell command, it's a `jq` operation (it's inside the `jq` format string), not a piped command. If you could elaborate on what you're seeing when you run the command, it'll help me troubleshoot differences for `jq` between Ubuntu and zbash so I can generalize the example code.

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
schavis: I know I copied the skeleton config from an example in the docs or tutorials but I can't find it now. I'll update to the `leader` syntax. Thanks for catching that. I wonder why my local Vault didn't complain 🤔

content/vault/v2.x/content/docs/deploy/run-on-docker.mdx
schavis: I'm just removing all the proxy references for now. Since this is already a pretty long doc, I'd like to have a follow-up guide specifically for adding a proxy to a Vault cluster (assuming we'd support something other than ZCx =\

PR #2579, opened 2026-06-04
Forward port context bug fix

williamdalessandro:
This PR fixes some bugs that currently exist in the forward port feature and also adds in some extra usage checks with better error messages.

- Fixes an issue where the script files would be lost when used on older PRs that don't have this script/feature in their git history due to the `Check out at merge commit` step
- added surface level checks of the to give better error messages during the JSON override manual dispatch path
  - checks for valid JSON
  - early check if all fields are present


RubenSandwich (approved):
👍🏻

PR #2584, opened 2026-06-04
Forward port code refactor

williamdalessandro:
Updates the files here to be wrapped as a module to make it easier to reference/reuse/test. Makes it less awkward for those files to live in this project compared to being used primarily as a script intended to only be ran in the command line.

Modifies/simplifies the associated tests accordingly


RubenSandwich (approved):
This looks great. ❤️ 

Love that these tests are much simpler now and using a mocked file system.

williamdalessandro:
> This looks great. ❤️
> 
> Love that these tests are much simpler now and using a mocked file system.

yeah the mocked file system is infinitely cleaner

PR #2654, opened 2026-06-16
RADAR-8728: Update documentation to include pre_commit_skip_activeness config option

Alan-Liang-Hashicorp:

### Terraform


content/hcp-docs/content/docs/vault-radar/cli/install/git.mdx
```diff
@@ -58,12 +58,15 @@ Note: The local verson of the configuration will have precedence over the global
...
-    "fail_severity": "high"
+    "fail_severity": "high",
+    "pre_commit_skip_activeness": true
 }
 ```
 
 </CodeBlockConfig>
 
 * `fail_severity` - This defines a fail threshold for vault-radar. When a risk is identified that has a severity at or beyond the configured fail-severity, `vault-radar` will consider the scan a failure. See here for more information on [severity](/hcp/docs/vault-radar/concepts/severity) and the different levels.
+* `pre_commit_skip_activeness` - `true` results in skipping activeness checks. The default is `true`. Enabling activeness checks will likely add more time to the evaluation, set to `false` with caution.
```
JoshAtHC: Shouldn't the default be `false`?

> Enabling activeness checks will likely add more time to the evaluation, set to `false` with caution.

I'm guessing there's some context bleeding in from the `pre-recieve` hook. That hook is special. It runs only on GitHub enterprise servers after someone has typed `git push` but before the server actually accepts the commit(s).

It has some quirks, such as it HAS to complete in less than 5 seconds. If it times out, the push is rejected regardless of whether it's safe or not. For that command we suggest disabling the activeness checks because they can add a lot of time and potentially make the whole hook execution exceed that timeout.

Outside of that scenario, there's no reason to disable the checks (other than a customer deciding they don't want them - which is what IBM is doing, and that's OK). We should encourage folks to leave them enabled and allow them to disable them as needed.

Alan-Liang-Hashicorp: Ah got it, thanks for the context! Have updated corresponding code to default false and will update this as well

content/hcp-docs/content/docs/vault-radar/cli/install/git.mdx
```diff
@@ -58,12 +58,14 @@ Note: The local verson of the configuration will have precedence over the global
 
 ```json
 {
-    "fail_severity": "high"
+    "fail_severity": "high",
+    "pre_commit_skip_activeness": true
```
jchen43: Nit: prob `false` since default val?

content/hcp-docs/content/docs/vault-radar/cli/install/git.mdx
```diff
@@ -58,12 +58,14 @@ Note: The local verson of the configuration will have precedence over the global
...
 }
 ```
 
 </CodeBlockConfig>
 
 * `fail_severity` - This defines a fail threshold for vault-radar. When a risk is identified that has a severity at or beyond the configured fail-severity, `vault-radar` will consider the scan a failure. See here for more information on [severity](/hcp/docs/vault-radar/concepts/severity) and the different levels.
+* `pre_commit_skip_activeness` - `true` results in skipping activeness checks. By default, activeness checks are enabled, i.e. the default value for this option is `false`
 
 **Note: If this configuration value is not defined the default behavior is to not enforce any severity. As a result all risks identified will be allowed.**
```
jchen43: Worth noting this is not accurate anymore since we have 2 config values now.

content/hcp-docs/content/docs/vault-radar/cli/install/git.mdx
Alan-Liang-Hashicorp: For this, I wanted to demonstrate when customer would want to skip activeness check by explicitly adding it in their config as an example. Otherwise, they would just omit it from config, and secret activeness checks would happen by default

content/hcp-docs/content/docs/vault-radar/cli/install/git.mdx
Alan-Liang-Hashicorp: Updated the Note section to be specific about fail_severity config value now

content/hcp-docs/content/docs/vault-radar/cli/install/git.mdx
jchen43: >Otherwise, they would just omit it from config, and secret activeness checks would happen by default

Isn't this traditional default behavior though? I personally feel like examples of configurations should typically be as close as default and any change in behavior should be understood through docs rather than boilerplate.

jchen43 (approved):

content/hcp-docs/content/docs/vault-radar/cli/install/git.mdx
Alan-Liang-Hashicorp: Got it, I see your point. Changed to false now

content/hcp-docs/content/docs/vault-radar/cli/install/git.mdx
```diff
@@ -58,12 +58,14 @@ Note: The local verson of the configuration will have precedence over the global
...
 ```
 
 </CodeBlockConfig>
 
 * `fail_severity` - This defines a fail threshold for vault-radar. When a risk is identified that has a severity at or beyond the configured fail-severity, `vault-radar` will consider the scan a failure. See here for more information on [severity](/hcp/docs/vault-radar/concepts/severity) and the different levels.
+* `pre_commit_skip_activeness` - `true` results in skipping activeness checks. By default, activeness checks are enabled, i.e. the default value for this option is `false`
 
-**Note: If this configuration value is not defined the default behavior is to not enforce any severity. As a result all risks identified will be allowed.**
\ No newline at end of file
+**Note: If `fail_severity` configuration value is not defined the default behavior is to not enforce any severity. As a result all risks identified will be allowed.**
```
jonathanfrappier: There are components for notes, warning, and highlight. This helps readers differentiate between UI elements (bold), code (either in ``` codeblocks or `code` format). Also, try to avoid passive voice by directly speaking to the reader


```suggestion
<Note> 

If you do not set the`fail_severity` configuration value, the default behavior is to not enforce any severity. As a result all risks identified will be allowed.

</Note>
```

jonathanfrappier (approved):

PR #2708, opened 2026-06-24
Compatibility matrix update 

paras-gupta2:
## Description

Updates the Consul on Kubernetes compatibility tables across `v1.21.x`, `v1.22.x`, and `v2.0.x`.

This PR refreshes the compatibility matrix partials under the versioned Consul docs to reflect the latest Kubernetes coverage and to make lifecycle state explicit in each table.

## What changed

- Updated Kubernetes compatibility ranges in the `version.mdx` and `lts.mdx` tables
- Updated cloud provider compatibility tables in `providers.mdx`
- Added a new `Lifecycle status` column to the Kubernetes compatibility tables
- Marked currently supported release lines as `Supported`
- Marked older release lines as `EOL`


anandmukul93:
why cant partials be common and updated in one place ?
this was those can be copied while website deployment  into every branch to render pages ?
can we sync with docs team to make this happen

paras-gupta2:
> why cant partials be common and updated in one place ? this was those can be copied while website deployment into every branch to render pages ? can we sync with docs team to make this happen

We can share partials in this repo, but these specific compatibility tables are not truly identical across versions. The content changes by release, so keeping a single common partial would either make some versioned pages inaccurate or require a larger refactor to redesign the tables as one shared cross-version matrix. For this PR, the version local partials are intentional because the compatibility data differs per release.

aimeeu (approved):

PR #2720, opened 2026-06-25
fix: adjust repo-sync permissions

LeahMarieBush:
### Description

This PR adds the permissions for the repo-sync action on the jobs. This is to fix an issue we are seeing where this action is not working properly


andys-hashi (approved):
lgtm

PR #2500, opened 2026-05-25
Api Gateway Dual GW Arch model documentation update

vdinesh4738:
Api Gateway Dual GW Arch model documentation update

PR #2770, opened 2026-07-02
resource upgrade flow

vdinesh4738:
resource upgrade flow for Consul 2.0, with the introduction of Dual Gateway feature


content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -127,8 +127,75 @@ global:
     enabled: true
 ```
 
+Identify the OpenShift version and determine the type of gateway.
```
Ranjandas: Considering this is the installation guide (and not upgrade guide), why don't we simply have the defaults with what we expect customers to run, so that we can avoid this conditional configurations? In a brand new installation, there won't be existing `tcproutes` that are managed by Consul.

If we are continuing with this, it is better if we include the command to do this.

Eg: `oc version`, `k get tcproute -A` or something along those lines.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -127,8 +127,75 @@ global:
...
+
+<Tabs>
```
Ranjandas, line 177: There are references to the term _"Standard Gateway"_ and _"Consul Gateway"_.  Where can the users find details on what these are? It will be confusing with our existing gateways like `API`, `Terminating`, `Mesh` etc. Shouldn't they be named _Gateway Controller_, instead of _Gateway_?

Ranjandas, line 177: The public Helm Chart reference don't seem to have any of this newly introduced options. Do we know when those are getting updated? Without which, it will be hard to understand what these changes are for.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -127,8 +127,75 @@ global:
...
+| OCP ≥ 4.19 | TCPRoute missing                             | Standard API Gateway | `installK8sNetworkingCrds=false`                                                                                 |
+| OCP ≥ 4.19 | TCPRoute present                             | Consul API Gateway   | `installK8sNetworkingCrds=false`, `openshift.crds.enableTcpRoute=false`, `openshift.crds.consulapi.enabled=true` |
+| Non-OCP    | Platform does not manage CRDs                | Standard API Gateway | `installK8sNetworkingCrds=true`                                                                                 |
+| Non-OCP    | Platform manages CRDs, TCPRoute not required | Standard API Gateway | `installK8sNetworkingCrds=false`                                                                                |
+| Non-OCP    | Platform manages CRDs, TCPRoute required     | Consul API Gateway   | `installK8sNetworkingCrds=false`, `crds.enableTcpRoute=false`, `crds.consulapi.enabled=true`                     |
+
```
Ranjandas: Same as above, these terms _Standard_ and _Consul_ API Gateway need to be revisited.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 layout: docs
-page_title: Resource upgrade
+page_title: New Gateway Installation and Upgrade Guide
```
Ranjandas: _Gateway Controller_ instead of _Gateway_?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+![Dual gateway new installation of Consul and OCP upgrade flow](/img/dual-gw-new-install-and-ocp-upgrade-flow.png)
+
+![Dual gateway non-OCP upgrade flow](/img/dual-gw-non-ocp-upgrade.png)
 
-    Moving to the new `consul.hashicorp.com` custom API resource group is not
-    required, but we do recommend it.
 
 ## Upgrade steps
 
+### OpenShift specific upgrade from 4.18 to 4.19+
```
Ranjandas: This section seems to be dealing with OpenShift specific upgrade workflow. Considering that it is a one-off thing, should it be a separate document rather than having inside the generic upgrade guide?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+
+1. Perform helm upgrade with below values in values.yaml.
```
Ranjandas: It is better if we explicitly call out that this is a pre-requisite before upgrading OCP, and also what all changes this step will bring to the resources, and also the steps to validate the outcome of this step before proceeding to the next.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+          enabled: true
+```
+
+</CodeBlockConfig>
+
+</Tab>
+
+</Tabs>
+
+2. Verify the access to the application.
```
Ranjandas: Having sample outputs highlighting two gateways running together will be useful in this step.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+   - **Standard Gateway**
+     - Verify the access to app with the gateway.
+
+   - **Consul Gateway**
+     - Users find 2 Gateways initially, try to access the app with both gateways.
+     - Update your DNS (e.g. Route53) to point to Consul DNS.
```
Ranjandas: What is this requirement? 🤔 Why should the DNS be pointed to Consul DNS? If this is a requirement mid upgrade, this should be called out in the pre-requisites.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+
+3. Copy the generated manifests from PVC as backup.
```
Ranjandas: Let us make sure we explain what these generated manifests are, how they came into existence and what is the use of it.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+
+4. Follow the official OpenShift documents to perform OCP platform upgrade which involves:
+   - deletion of Gateway CRDs.
+   - Approving adminAck and so on
+
+5. Verify the access to the application after platform upgrade.
+   - **Standard Gateway**
+     - apply the manifests copied.
```
Ranjandas: Lets explain things a bit more here. I guess this is referring to the manifests copied in Step 3?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+     - Once gateway is up, try to access the application.
+    
+   - **Consul Gateway**
+     - Try to access the application.
+     - if access to application is fails, perform helm upgrade with:
```
Ranjandas: What is the condition where access to the application would fail? Is it not possible to pre-determine this condition and avoid it?

Can we explain what the issue is and what we doing differently to fix the issue?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+       openshift:
+          enabled: true
+          upgradeTo419From418: false
+          crds:
+            enableTcpRoute: false
+              consulapi:
+              enabled: true
+        ```
+ 
+### Non-OCP upgrade flow : Platform upgrade not involved.
```
Ranjandas: Should this page be dedicated to K8S/OCP, as if it was not for the 4.18 to 4.19+ upgrade for OCP, the steps would be same for both OCP and K8S right? So we can simplify this whole doc by separating the two?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+        enableTcpRoute: true
+```
+
+</CodeBlockConfig>
+
+</Tab>
+
+</Tabs>
+
+2. Access the Application
```
Ranjandas: Let us have more details here.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -33,13 +30,212 @@ Kubernetes clusters not using the `TCPRoute` resource can, but do not need to, u
...
+
+2. Verify access to the application
+   - **Standard Gateway**
+     - Verify the access to app with the gateway.
+   - **Consul Gateway**
+     - Users find 2 Gateways initially, try to access the app with both gateways.
+     - Update your DNS (e.g. Route53) to point to Consul DNS.
+     - Delete the Standard Gateway and its objects.
+
+3. Copy the generated manifests.
```
Ranjandas: What are we going to do with these manifests?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -63,16 +259,7 @@ To upgrade your Kubernetes resources for Consul, follow these steps:
...
-
-   Run the following command to delete the old CRDs:
-
-    ```bash
-    for i in gateways gatewayclasses httproutes grpcroutes tcproutes referencegrants; \
-    do oc delete crd $i.gateway.networking.k8s.io; \
-    done
-    ```
+<Note>
+ Make backups of all the manifests after the Consul upgrade or copy the manifests from PVC before proceeding to the OCP upgrade.
```
Ranjandas: Is this step misplaced?

Ranjandas (commented):
High-level review of the steps. I think we should refine this documentation as we will be having customers upgrading and will be looking for detailed upgrade guide. 

It is important to call out what each steps outcome are, with some example commands/outputs.

The document also don't have explicit mention of whether the upgrade steps involved downtime or not. 

There are no rollback plans mentioned in the doc. 

There are references to terms like "Consul Gateway/Standard Gateway", without having explanation on what those are.

There are helm values that are not yet available in the public Helm Chart reference. 

Lets also make sure that the YAML indentation is consistent across the doc.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
vdinesh4738: in brand new installation, currently without this feature we actually watch for the TCPRoute by default, whether or not the customer is having an application with the TCPRoute. Thus we need to know if the Customers have a TCP Application with TCPRoute under pre-requisites and based on the requirement we watch for TCPRoute or not

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
vdinesh4738: This is defined in the introduction page in another PR I have shared. Gateway is a resource and Gateway Controller reconcile it. We can not name it

vdinesh4738: https://artifacthub.io/packages/helm/hashicorp/consul
The above link has the helm values file with the options.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
vdinesh4738: It has both generic and openshift specific. I am not sure if you want me to create a OCP specific page.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
vdinesh4738: @Ranjandas Can you please tell me how this looks - Standard API Gateway and Consul API Gateway?

vdinesh4738: I added those values now

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -127,8 +127,75 @@ global:
...
+
 Refer to [`openshift` in the Helm chart reference](/consul/docs/reference/k8s/helm#v-global-openshift) for additional information.  
 
+### Installation Decision Summary
+
+| Platform   | Condition                                    | Gateway              | Helm Values                                                                                                                                 |
+| ---------- | -------------------------------------------- | -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
+| OCP < 4.19 | Always                                       | Standard API Gateway | `installK8sNetworkingCrds=true`, `openshift.crds.enableTcpRoute=true`                                            |
+| OCP ≥ 4.19 | TCPRoute missing                             | Standard API Gateway | `installK8sNetworkingCrds=false`                                                                                 |
```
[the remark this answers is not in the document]
danielehc: @vdinesh4738 the automated check noticed that in the table we use `installK8sNetworkingCrds` while in the examples we use `installK8sNetworkingCRDs`.

Are they bot allowed? if not which one is the correct one?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -16,14 +16,11 @@ Kubernetes clusters. Consul 2.0.0 provides new resource
...
-to replace the older resource types that previous versions of Consul used.
+to support the TCPRoute functionality in the Kubernetes platforms where Gateway CRDs are managed.
 
 ## Requirements
 
-- OpenShift Container Platform (OCP) 4.19 or later
-- Consul 2.0.0 or later
-
-Update the API gateway resources from the current API resource group `gateway.networking.k8s.io` to the new `consul.hashicorp.com`. This change is necessary to take advantage of the new features and improvements in the API gateway functionality on specific Kubernetes environments, like OpenShift.
+- Consul deployed in the Kubernetes.
```
danielehc:
```suggestion
- Consul deployed in Kubernetes.
```

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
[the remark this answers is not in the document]
vdinesh4738: resolved it. The correct one is installK8sNetworkingCRDs

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -684,14 +686,70 @@ Use these links to navigate to a particular top-level stanza.
...
+  - `generateManifests` ((#v-global-generatemanifests)) (`boolean: false`) - Set to `true` to generate the manifests. After you generate the manifests, set this to `false` for subsequent Helm upgrades.
+
+  - `crds` ((#v-global-crds)) - During the install:
+    custom crds and gateway TCP crd enabling
+    enableTcpRoute installs the TCP GVK under gateway.networking.k8s.io and enables the stable gatewaycontroller to watch for the resources.
+    crds.consulapi.enabled install the CRDs under consul.hashicorp.com API group and enables the new custom gateway controller to watch for the resources.
+
+    During the upgrade - manifests are generated.
+    - If consulapi.enabled is set to `false`, the chart does not generate manifests under the `consul.hashicorp.com` API group.
+    - If consulapi.enabled is set to `true`; the chart generates and applies manifests under the `consul.hashicorp.com` API group.
```
danielehc:
```suggestion
  - `crds` ((#v-global-crds)) - This flag has different behaviors during the installation process and during the upgrade process.
   
    During the installation, custom CRDs and gateway TCP CRD are enabled.
    - `enableTcpRoute` installs the TCP GVK under `gateway.networking.k8s.io` and enables the stable gateway controller to watch for the resources.
    - `crds.consulapi.enabled` installs the CRDs under `consul.hashicorp.com` API group and enables the new custom gateway controller to watch for the resources.

    During the upgrade, manifests are generated.
    - If `consulapi.enabled` is set to `false`, the chart does not generate manifests under the `consul.hashicorp.com` API group.
    - If `consulapi.enabled` is set to `true`, the chart generates and applies manifests under the `consul.hashicorp.com` API group.
```

We refer to `crds.consulapi.enabled` in the installation and to `consulapi.enabled` during the upgrade, are they the same? If so we should use one name for both.

danielehc: ⚠️ After applying the change please double check if the preview is visualized correctly.

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -684,14 +686,70 @@ Use these links to navigate to a particular top-level stanza.
...
+    - `upgradeTo419From418` ((#v-global-openshift-upgradeto419from418)) (`boolean: false`) - During the install:
+      in this case ideally customer deploys the consul directly on the 4.19+ and thus is void.
+
+      During the upgrade:
+      This should be enabled when the customer plans to upgrade the OCP from 4.18 to 4.19+ versions.
+      This will generate the manifests inaccordance to 2 cases along with enableTcpRoute:
+      case 1: when upgradeTo419From418 is true and enableTcpRoute is false:
+      generate the manifests only for gateway.networking.k8s.io and ignores the TCP watch
+      case 2: when upgradeTo419From418 is true and enableTcpRoute is true:
+      generate the manifests for both consul.hashicorp.com and gateway.networking.k8s.io ( the deletion of TCP GVK is manual )
```
danielehc:
```suggestion
    - `upgradeTo419From418` ((#v-global-openshift-upgradeto419from418)) (`boolean: false`) - This flag has different behaviors during the installation process and during the upgrade process. 
      During the installation: if the installation is performed on an OCP instance 4.19+ the flag is ignored. We recommend to perform new installations on an OCP instance 4.19+.

      During the upgrade: This flag should be enabled when upgrading the OCP from 4.18 to 4.19+ versions.
      When enabled, it will generate the manifests. The resources in the manifest depend on the value of `enableTcpRoute`:
      - case 1: when `upgradeTo419From418` is `true` and `enableTcpRoute` is `false`, the upgrade generates the manifests only for `gateway.networking.k8s.io` and ignores the TCP watch.
      - case 2: when `upgradeTo419From418` is `true` and `enableTcpRoute` is `true`, the upgrade generates the manifests for both `consul.hashicorp.com` and `gateway.networking.k8s.io` ( this process requeres the manual deletion of TCP GVK )
```

danielehc: ⚠️ After applying the change please double check if the preview is visualized correctly.

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -684,14 +686,70 @@ Use these links to navigate to a particular top-level stanza.
...
+
+      in any case, crds.consulapi.enabled install the CRDs under consul.hashicorp.com API group and enables the new custom gateway controller to watch for the resources.
+
+    - `crds` ((#v-global-openshift-crds))
+
+      - `enableTcpRoute` ((#v-global-openshift-crds-enabletcproute)) (`boolean: false`)
+
+      - `consulapi` ((#v-global-openshift-crds-consulapi))
+
+        - `enabled` ((#v-global-openshift-crds-consulapi-enabled)) (`boolean: false`)
```
danielehc:
```suggestion
    - `isOcpGreaterthan4_18` ((#v-global-openshift-isocpgreaterthan4_18)) (`boolean: false`) - This flag has different behaviors during the installation process and during the upgrade process.  
      During the installation,  if the installation is performed on an OCP instance 4.19+ the flag is ignored. We recommend to perform new installations on an OCP instance 4.19+.
      - If `crds.enableTcpRoute` is `true`, the standard gateway controller watches `TCPRoute` resources.
      - If `crds.enableTcpRoute` is `false`, the standard gateway controller does not watch TCPRoute resources.
      - if `crds.consulapi.enabled` is `true`, controller watches for resources under the API group `consul.hashicorp.com`.

      During the upgrade:
      - If `isOcpGreaterthan4_18` is `false` and `enableTcpRoute` is `false`, generates the manifests only for `gateway.networking.k8s.io` and ignores the TCP watch.
      - If `isOcpGreaterthan4_18` is `false` and `enableTcpRoute` is `true`, generates the manifests only for `gateway.networking.k8s.io` and watches for the TCP.
      - If `isOcpGreaterthan4_18` is `true` and `enableTcpRoute` is `false`, generates the manifests only for `gateway.networking.k8s.io` and ignores the TCP watch
      - If `isOcpGreaterthan4_18` is true and `enableTcpRoute` is true, generates the manifests for both `consul.hashicorp.com` ( all current GVK ) and `gateway.networking.k8s.io` ( no TCP )
      - In any case, `crds.consulapi.enabled` installs the CRDs under `consul.hashicorp.com` API group and enables the new custom gateway controller to watch for the resources.

    - `crds` ((#v-global-openshift-crds))

      - `enableTcpRoute` ((#v-global-openshift-crds-enabletcproute)) (`boolean: false`)

      - `consulapi` ((#v-global-openshift-crds-consulapi))

        - `enabled` ((#v-global-openshift-crds-consulapi-enabled)) (`boolean: false`)
```

We refer to `crds.enableTcpRoute` in the installation and to `enableTcpRoute` during the upgrade, are they the same? If so we should use one name for both.

> watches for the TCP
Could you rephrase this to make it clearer? Do you mean "whatches for the TCPRoute resourcers"?

danielehc: ⚠️ After applying the change please double check if the preview is visualized correctly.

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -2019,13 +2077,28 @@ Use these links to navigate to a particular top-level stanza.
 
       - `resources` ((#v-connectinject-apigateway-managedgatewayclass-resources)) (`map`) - The resource settings for Pods handling traffic for Gateway API.
 
-      - `deployment` ((#v-connectinject-apigateway-managedgatewayclass-deployment)) - This value defines the number of pods to deploy for each Gateway as well as a min and max number of pods for all Gateways
+      - `scaling` ((#v-connectinject-apigateway-managedgatewayclass-scaling)) - <EnterpriseAlert inline /> These settings enable per-Gateway scaling behavior for
+        the managed GatewayClass, including Gateway annotations for static
+        replicas, controller-managed HPA resources, and preserving manual scale.
+        Requires Consul Enterprise with a valid license.
+
+        - `enabled` ((#v-connectinject-apigateway-managedgatewayclass-scaling-enabled)) (`boolean: false`)
```
danielehc: Could you add the description for this field?

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -2019,13 +2077,28 @@ Use these links to navigate to a particular top-level stanza.
...
 
-        - `defaultInstances` ((#v-connectinject-apigateway-managedgatewayclass-deployment-defaultinstances)) (`integer: 1`)
+      - `deployment` ((#v-connectinject-apigateway-managedgatewayclass-deployment)) - DEPRECATED: These values are deprecated and will be removed in a future release.
+        Use Gateway annotations instead for per-gateway scaling configuration.
+        See: [`API gateway scaling`](/consul/docs/k8s/api-gateway/scaling).
```
danielehc: The link does not work in preview.

Do you mean /consul/docs/north-south/api-gateway/k8s/scale ?

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -2019,13 +2077,28 @@ Use these links to navigate to a particular top-level stanza.
...
 
-        - `minInstances` ((#v-connectinject-apigateway-managedgatewayclass-deployment-mininstances)) (`integer: 1`)
+        - `defaultInstances` ((#v-connectinject-apigateway-managedgatewayclass-deployment-defaultinstances)) (`integer: 1`) - Default number of gateway instances
+          DEPRECATED: Use consul.hashicorp.com/default-replicas annotation on Gateway resource
+
+        - `maxInstances` ((#v-connectinject-apigateway-managedgatewayclass-deployment-maxinstances)) (`integer: 1`) - Maximum number of gateway instances
+          DEPRECATED: No longer enforced. Use HPA maxReplicas via annotations instead
+
+        - `minInstances` ((#v-connectinject-apigateway-managedgatewayclass-deployment-mininstances)) (`integer: 1`) - Minimum number of gateway instances
+          DEPRECATED: Use HPA minReplicas via annotations instead
```
danielehc: Use `monospace` text for parameters.

Where are `consul.hashicorp.com/default-replicas`, `HPA maxReplicas`, `HPA minReplicas` defined. Please link to them.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,302 @@
...
+
+```shell-session
+$ mkdir -p gateway-resource-backup && \
+  for ns in $(oc get ns -o jsonpath='{.items[*].metadata.name}'); do \
+    oc get gateways -n "$ns" -o yaml > "gateway-resource-backup/${ns}-gateways.yaml"; \
+    oc get httproutes -n "$ns" -o yaml > "gateway-resource-backup/${ns}-httproutes.yaml"; \
+    oc get tcproutes -n "$ns" -o yaml > "gateway-resource-backup/${ns}-tcproutes.yaml"; \
+    oc get referencegrants -n "$ns" -o yaml > "gateway-resource-backup/${ns}-referencegrants.yaml"; \
+  done
+```
```
danielehc: Is this required only for OCP?
* If so make it an h3 (`###`) and name it OCP specific requirements (or something along this line)
* If not provide the equivalent command for non OCP k8s and mke it a tab for OCP and non OCP, still under an H3.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,302 @@
...
+
+## Additional Info
```
danielehc: This should be an H3 (`###`)

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,302 @@
...
+If your cluster does not use `TCPRoute`, you can keep existing gateway resources.
+If your cluster uses `TCPRoute`, upgrade to the new API group in one of these cases:
+
+- OCP clusters that use `TCPRoute` and upgrade from OCP 4.18 to 4.19 or later.
+
+  Move to the `consul.hashicorp.com` API group to maintain compatibility with OCP 4.19 and later.
+
+- Non-OCP Kubernetes clusters that use `TCPRoute` and platform-managed Gateway CRDs.
+
+  Move to the `consul.hashicorp.com` API group to maintain compatibility.
```
danielehc, line 85: This one should be a flowchart. The initial step to understand where I stand as a customer, and which steps are required.

Image

This can be used as a base to understand what I mean.

Also maybe, we should explain how to understand if we use TCPRoute or not, is there a command for it?

From this we can now have two H2 (`##`):
* Kubernetes upgrade steps
* OpenShift Container Platform upgrade steps

The list items will link to the sections. You can also invert the two.

In **Kubernetes upgrade steps:**

Start with the flowchart that explains the different conditions:

Image

Then have the two sections:

* Kubernetes upgrade flow: platform upgrade not involved (H3 `###`)
* Kubernetes upgrade flow: platform upgrade involved (H3 `###`)

In **OpenShift Container Platform upgrade steps:**

Start with the **OCP upgrade requirements and resource involved.** section (H3 `###`)
Then show the flowchart:

Image

Then add the **OpenShift upgrade from 4.18 to 4.19+** section  (H3 `###`)

danielehc: Visually it should look like:

```
## Kubernetes upgrade steps
     --> Flowchart NON-OCP upgrade flow
### Kubernetes upgrade flow: platform upgrade not involved 
### Kubernetes upgrade flow: platform upgrade involved (H3 ###)
## OpenShift Container Platform upgrade steps
     --> Flowchart OCP upgrade flow
### OCP upgrade requirements and resource involved
### OpenShift upgrade from 4.18 to 4.19+

```

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,302 @@
...
+
+## Upgrade flow charts
+
+Use these flow charts to choose the correct path for upgrade and new installation.
+
+1. New installation of Consul
+
+![New installation of Consul](/img/new-installation.png)
```
danielehc: This flowchart would make much more sense in the installation page.

/consul/docs/deploy/server/k8s/helm

danielehc (changes requested):
Left some comments:

* The reference render is incorrect so I suggested some fixes to make it render properly. 
* The deploy page is basically there but I suggested we should add the installation flowchart there, near the table to help customers visualize requirements
* The upgrade page is still a bit confusing, suggested some structure changes to make the flow easier to understand and to better leverage the flowcharts

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -684,14 +686,70 @@ Use these links to navigate to a particular top-level stanza.
   - `imageConsulDataplane` ((#v-global-imageconsuldataplane)) (`string: hashicorp/consul-dataplane:<latest supported version>`) - The name (and tag) of the consul-dataplane Docker image used for the
     connect-injected sidecar proxies and mesh, terminating, and ingress gateways.
 
+  - `generateManifests` ((#v-global-generatemanifests)) (`boolean: false`) - Set to `true` to generate the manifests. After you generate the manifests, set this to `false` for subsequent Helm upgrades.
```
Ranjandas: Should we link to the upgrade guide for context of what this is? Especially being a one-off use flag.

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -2335,6 +2409,18 @@ Use these links to navigate to a particular top-level stanza.
...
+    RateLimit).
+
+    This token is optional and only used when both secretName and secretKey
+    are provided. When set, connect-inject passes only the secret name/key
+    and reads the token from the Kubernetes Secret in code.
+
+    - `secretName` ((#v-connectinject-globalconfigacltoken-secretname)) (`string: null`) - The name of the Kubernetes secret that holds the global config ACL token.
+
+    - `secretKey` ((#v-connectinject-globalconfigacltoken-secretkey)) (`string: null`) - The key within the Kubernetes secret that holds the global config ACL token.
+
```
Ranjandas, line 2417: Is there documentation on what this new feature is about? And should that be linked here for reference?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,302 @@
...
 ## Overview
 
-Kubernetes resources for Consul provide a way to configure and manage Consul on
-Kubernetes clusters. Consul 2.0.0 provides new resource
-types that offer enhanced functionality and improved compatibility
-with OpenShift Container Platform (OCP) versions 4.19 and above. These new
-resource types are part of the `consul.hashicorp.com` API group and are designed
-to replace the older resource types that previous versions of Consul used.
+Consul 2.0.0 introduces API gateway resource types in the `consul.hashicorp.com`
+API group. These resource types improve compatibility with OpenShift Container
```
Ranjandas, line 19: What additional compatibility is this API Group providing "specifically" OCP? 🤔 Is that statement accurate? From my understanding this API Group is only there to support TCPRoute. Isn't that the case?

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,302 @@
...
+    oc get httproutes -n "$ns" -o yaml > "gateway-resource-backup/${ns}-httproutes.yaml"; \
+    oc get tcproutes -n "$ns" -o yaml > "gateway-resource-backup/${ns}-tcproutes.yaml"; \
+    oc get referencegrants -n "$ns" -o yaml > "gateway-resource-backup/${ns}-referencegrants.yaml"; \
+  done
+```
+
+## Additional Info
+- If the customers have the GoDaddy or Route53 as the domain providers and have their CNAME pointed to the API Gateway, they should update their CNAME on moving to the Consul API Gateway from the Standard API Gateway.
+- As part of upgrading to Consul 2.0, Consul generates manifests and update them to v1 version of the `gateway.networking.k8s.io` incase customer continues to use `Standard API Gateway` which is the recommended API Gateway and applies them.
+  Similarly, when the customer selects the `Consul API Gateway`, Consul generates manifests for both Standard API Gateway and Consul API Gateway and applies them, thus creating both API Gateways. This is done to provide the flexibitliy for the Customer to update CNAME in their Domain provider ( ex: GoDaday or Route53 if applicable).
```
Ranjandas: it reads more like internal engineering notes than end-user documentation. feels third-person instructional writing.

Ranjandas (commented):
I have reviewed and feels like this still needs improvement. Some of the review comments I added in the previous commit was not addressed and I can see the same mistakes in the latest changes.

Please note that I won't have bandwidth to do the reviews going forward. 

I would highly recommend to run the content of this PR through Bob and make it more refined.

content/consul/v2.0.x/img/new-installation.png
himanshu-8693, line 1: The OCP ≤4.18 path shows `openshift.crds.enableTcpRoute: true` as a required value. From some support case work and chart source, `enableTcpRoute` defaults to `false` on the OCP path and the chart comment says TCPRoute on OCP ≤4.18 is fine with true — but it is not required for a basic Standard API Gateway install on OCP ≤4.18. The diagram implies it is mandatory. Minor but could confuse customers who don't use TCPRoute.

content/consul/v2.0.x/img/ocp-upgrade.png
himanshu-8693, line 1: The diagram title and entry gate say "Is current version 4.18 and upgrading to 4.19 or 4.19+?" with a hard No exit saying "This flow is applicable only for upgrades from 4.18 to 4.19+". 
Does it means a customer going from some other OCP version eg. 4.16 → 4.20 would be told this diagram doesn't apply to them and get no guidance ?

himanshu-8693, line 1: For the OCP upgrade >=4.19 if TCP route is required and Consul API GW is needed, diagram still set `openshift.crds.enableTcpRoute: false`, shouldn't it be `true` instead ?

himanshu-8693, line 1: `installK8sNetworkingCRDs: true` in both pre-upgrade Helm value boxes. This is shown without any warning. On basis of some support cases, setting this true during the Consul Helm upgrade (while still on old OCP) stamps `manager: helm` on the gateway CRDs — creating a mandatory cleanup step before the OCP upgrade. The diagram shows no cleanup step and no mention of `manager: helm`.

"Delete Gateway API CRDs (if required)" is mentioned as a single bullet under "Follow OCP upgrade documents" — but gives no detail on the `storedVersions` migration, the ValidatingAdmissionPolicy blocking deletion, or the orphaned CRD (tcproutes/tlsroutes/udproutes) problem. These are not optional — they are blockers.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
himanshu-8693, line 1: Just a few observations:-

* OCP ≥4.19 Standard API Gateway tab is missing `isOcpGreaterthan4_18: true` — without it the chart doesn't switch to the correct CRD path
* Decision table lists `enableTcpRoute=true` as always-required for OCP ≤4.18 — it's only needed if the workload actually uses TCPRoute

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: This is not OCP specific .This API gateway supports all routes including TCP route in Platforms that manage gateway CRDs ( OCP and GKE AutoPilot).

The sentence already has the details (line 20-21)

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: added a Note for clarification.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: done.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: Reframed the sentences.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: moved.

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
Manishakumari-hc: Done

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
Manishakumari-hc: corrected.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
danielehc: The link in the warning box does not render properly.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -39,6 +39,10 @@ Using the Helm Chart requires Helm version 3.6+. Visit the [Helm website](https:
```
danielehc:
```suggestion
You must meet the following requirements to install Consul on Kubernetes with Helm:

- Helm version `3.6+`. Visit the [Helm website](https://helm.sh/docs/intro/install/) to download the latest version.

- [`kubectl`](https://kubernetes.io/docs/reference/kubectl/) configured to point to the Kubernetes cluster you want to install Consul on

```

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -39,6 +39,10 @@ Using the Helm Chart requires Helm version 3.6+. Visit the [Helm website](https:
 
 ## Install Consul
```
danielehc, line 42:
```suggestion

Depending on the requirements you can:

- Perform a [default installation](#link_to_section) by using the default Helm chart configuration. Use this option if you are testing Consul or if you need to setup a test environment that does not have security requirements.
- Perform a [custom installation](#link_to_section) by modifying Helm chart configuration parameters. This is the recommended approach if you are installing Consul in production.
 
```

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -39,6 +39,10 @@ Using the Helm Chart requires Helm version 3.6+. Visit the [Helm website](https:
...
 
+### Flowchart for new installation of consul
+
+![New installation of Consul](/img/new-installation.png)
+
```
danielehc: This one is not referring to the whole installation, especially not to the default installation.

We should move it into the proper section.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -39,6 +39,10 @@ Using the Helm Chart requires Helm version 3.6+. Visit the [Helm website](https:
...
 1. Add the HashiCorp Helm Repository:
```
danielehc: This is required in both standard and custom installation, it might be moved into prerequisites or in its own H2 section.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
danielehc: This section could be its own H2 as last section.

danielehc, line 108:
```suggestion

This paragraph contains some customization you might want to use as reference:

- [Minimal Consul service mesh installation](#minimal-values-yaml-for-consul-service-mesh).
- [Select API gateway configuration](#add_relevant section)
- [Install Consul on OpenShift clusters](#install-consul-on-openshift-clusters).
- [Enable Consul CNI plugin](#enable-the-consul-cni-plugin).
- [Enable Consul service mesh on selected namespaces](#enable-consul-service-mesh-on-select-namespaces).
- 

```

Double check the link after changing the H3 section titles

danielehc, line 137: Since we have a diagram and a table that explain configuration for API gateway and then an OpenShift specific section to configure Consul that depends on the API gateway choice I would add a specific section to select the API gateway configuration.


```suggestion

### Select API gateway configuration

Here I will add a paragraph that explains that, since a fully fledged Consul deployment requires an API gateway, it is important to identify your platform to choose the best API gateway implementation for your infrastructure.

After that I would add the diagram, saying something like "The following is a diagram to help you pick the right configuration".

<DIAGRAM>

After that I would add the table, saying something like "The table shows the same information than the diagram explaining the parameter values"

<TABLE>

```

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -127,8 +131,87 @@ global:
...
+Identify the OpenShift version, then confirm whether the application requires `TCPRoute` to determine which API gateway to use.
+
+
+```shell-session
+$ oc version
+```
+
+```shell-session
+$ kubectl get tcproutes -A
+```
```
danielehc: Can we get some guidance on how to interpret the output of the commands above?

Smt like, if the command returns <example output> then use X otherwise do Y

Then point to the section before (the one containing all possible choice paths) that explains how to identify the right choice if they did not yet.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -127,8 +131,87 @@ global:
...
+  </Tab>
+  
+</Tabs>
+
+
 Refer to [`openshift` in the Helm chart reference](/consul/docs/reference/k8s/helm#v-global-openshift) for additional information.  
 
+### Installation Decision Summary
+
+| Platform   | Condition                                    | Gateway              | Helm Values                                                                                                                    |
```
danielehc: What is `Condition` referring to?

From the values possible conditions seem to be:
* When is `?????` required?
* Is TCPRoute present/required?
* Does platform manage CDRs?

I would replace `Condition` with `Platform configuration`

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -127,8 +131,87 @@ global:
...
+| ---------- | -------------------------------------------- | -------------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| OCP < 4.19 | Always                                       | Standard API Gateway | `global.installK8sNetworkingCRDs=true`, `global.openshift.crds.enableTcpRoute=true`                                            |
```
danielehc: I would replace `Always` with `OpenShift version < 4.19`.

Even if this information is already present in the first column, that column is named "Platform" so it is not immediate that that information is also a "Condition"

content/consul/v2.0.x/img/new-installation.png
danielehc, line 1: Some comments on the diagram:

* **Naming**: Before this PR we (almost) always used `API gateway` with lowercase g to define the feature. In this PR we move between `API Gateway` and `API gateway`. I think we should adopt the standard naming API gateway across all documentation, including this PR. Once we take the decision we should arrange the naming in the diagram as well.

* **Color scheme**: In the legend you have the blue clolor that seems used for the legend square itself but is also used to define outcome 1. At the same time the first option (OCP < 4.18) is coloured in green but has no number. I think 1st and 4th options should have the same color since they are basically the same outcome, if not, modify the diagram to add an appropriate number and color to all possible outomes.
 
* **Flow arrows and notes**: the note on the left `No (< 4.18)` should be `No (<= 4.18)` according to what we say in the table and in the tab dropdown. Also a minor thing would be that the color code mentioned in the legend is not always respected by the arrows.

* **Scenario configuration options**: the diagram offers also the configuration required by each outcome. I noticed that `generateManifests` is always false and that it is not present in the examples provided by the configuration. This seems correct since the option is reserved to the Upgrade process but I think it might then be removed from this diagram.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
danielehc: Also in the diagram and in the dropdown we use `OCP <= 4.18` and in this table `OCP < 4.19`. While the two options are logically equivalent it is confusing to have the same option named in two different ways in different sections on the page.

Decide on one and change the other instances in the document.

I am partial to `OCP <= 4.18` for two reasons:
* It requires less changes
* It provides a clearer distinction between the two versions that does not require to re read or notice small differences in the comparative symbols.

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -46,6 +46,8 @@ Use these links to navigate to a particular top-level stanza.
 
 - `global` ((#v-global)) - Holds values that affect multiple components of the chart.
 
+  - `installK8sNetworkingCRDs` ((#v-global-installk8snetworkingcrds)) (`boolean: true`) - Enable this flag to install the CRDs under `gateway.networking.k8s.io` API group. You can set this value to false if CRDs are externally installed before Consul installation or if CRDs are managed by the platform.
```
danielehc:
```suggestion
  - `installK8sNetworkingCRDs` ((#v-global-installk8snetworkingcrds)) (`boolean: true`) - Enable this flag to install the CRDs under `gateway.networking.k8s.io` API group. You can set this value to `false` if CRDs are externally installed before Consul installation or if CRDs are managed by the platform.
```

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
```diff
@@ -684,14 +686,72 @@ Use these links to navigate to a particular top-level stanza.
...
   - `openshift` ((#v-global-openshift)) - Configuration for running this Helm chart on the Red Hat OpenShift platform.
     This Helm chart currently supports OpenShift v4.x+.
 
     - `enabled` ((#v-global-openshift-enabled)) (`boolean: false`) - If true, the Helm chart will create necessary configuration for running
       its components on OpenShift.
 
-    - `upgradeTo419From418` ((#v-global-openshift-upgradeto419from418)) (`boolean: false`) - If true, the Helm chart upgrades existing resources to be compatible with OpenShift v4.19+.
-      Only set to true if you are currently running on OpenShift v4.18 and plan to upgrade to v4.19 imminently.
+    - `upgradeTo419From418` ((#v-global-openshift-upgradeto419from418)) (`boolean: false`) - This flag has different behaviors during the installation process and during the upgrade process. 
+      During the installation: if the installation is performed on an OCP instance 4.19+ the flag is ignored. We recommend to perform new installations on an OCP instance 4.19+.
```
danielehc:
```suggestion
      
      During the installation: if the installation is performed on an OCP instance 4.19+ the flag is ignored. We recommend to perform new installations on an OCP instance 4.19+.
```

This should be separated from the previous paragraph.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,311 @@
...
+    oc get gateways -n "$ns" -o yaml > "gateway-resource-backup/${ns}-gateways.yaml"; \
+    oc get httproutes -n "$ns" -o yaml > "gateway-resource-backup/${ns}-httproutes.yaml"; \
+    oc get tcproutes -n "$ns" -o yaml > "gateway-resource-backup/${ns}-tcproutes.yaml"; \
+    oc get referencegrants -n "$ns" -o yaml > "gateway-resource-backup/${ns}-referencegrants.yaml"; \
+  done
+```
+
+<Note>
+
+For OpenShift, use the `oc` client. For all other platforms, use `kubectl`.
```
danielehc: Instead of a note section this should be a dropdown that provides also a `kubectl` command to execute the same operation on non OCP clusters.

Also `kubectl` should be listed as a prerequisite for non-OCP clusters

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,311 @@
...
+
+- `Gateway`
+- `HTTPRoute`
+- `TCPRoute`
+- `ReferenceGrant`
 
 ## Upgrade steps
 
-To upgrade your Kubernetes resources for Consul, follow these steps:
+### OpenShift upgrade from 4.18 to 4.19+
```
danielehc, line 126: Numbers and bullet points are not rendering properly. Please fix numbering and indentation.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,311 @@
...
+          upgradeTo419From418: false
+          crds:
             enableTcpRoute: false
             consulapi:
-                enabled: true
-    ```
+              enabled: true
+      ```
+ 
+### Non-OCP upgrade flow: platform upgrade not involved
```
danielehc, line 225: Numbers and bullet points are not rendering properly. Please fix numbering and indentation.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,311 @@
...
+</Tab>
+
+</Tabs>
+
+2. Verify application access.
+  - Access the application using the the Standard API Gateway 
+
+3. Back up generated manifests from the PVC.
+
+### Non-OCP upgrade flow: platform upgrade involved
```
danielehc, line 271: Numbers and bullet points are not rendering properly. Please fix numbering and indentation.

danielehc (changes requested):
Left some more comments with change suggestions.

One thing I also mentioned in #2500  

> Before this PR we (almost) always used `API gateway` with lowercase g to define the feature. In this PR we move between `API Gateway` and `API gateway`. I think we should adopt the standard naming `API gateway` across all documentation, including this PR. Once we take the decision we should arrange the naming in the diagram as well.

content/consul/v2.0.x/img/ocp-upgrade.png
Manishakumari-hc: no it should be false.

Manishakumari-hc: Added "Validate if the gateway CRDs under API group `gateway.networking.k8s.io` are pre-installed in the cluster.If pre-installed, set `global.installK8sNetworkingCRDs: false` "  in resource-upgrade.mdx file.

"Delete Gateway API CRDs (if required)" is mentioned as a single bullet under "Follow OCP upgrade documents" — but gives no detail on the storedVersions migration, the ValidatingAdmissionPolicy blocking deletion, or the orphaned CRD (tcproutes/tlsroutes/udproutes) problem. These are not optional — they are blockers. ---> We are recommending to follow the official document of OCP.

Manishakumari-hc: addressed in image

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: addressed.

Manishakumari-hc: addressed.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
Manishakumari-hc: Moved to above requirement section

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
Manishakumari-hc: Added the rate limiting reference doc link.

content/consul/v2.0.x/img/new-installation.png
Manishakumari-hc: * Made the images black and white
* Added the ver <=4.18 in ocp-upgarde.png
* @danielehc captilalization issue in image "API Gateway" to "API gateway" still remains, reason being no tool is helping us rightly and it changes the image itself.
We have corrected this capiltalization throughout the code though.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: corrected as suggested.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: changed to "step x" concept and is rendering well.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: changed to "step x" concept and is rendering well.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: changed to "step x" concept and is rendering well.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,376 @@
...
+## Upgrade flow charts
+
+Use the following flow charts to choose the correct path for your upgrade.
+
+### Kubernetes upgrade flow: platform upgrade not involved 
+
+![platform upgrade not involved flow](/img/non_ocp_upgrade.png)
+
+### Kubernetes upgrade flow: platform upgrade involved (H3 ###)
+![platform upgrade involved flow](/img/ocp_platform_upgrade.png.png)
```
aimeeu: This doesn't render 

Image

```suggestion
### Kubernetes upgrade flow: platform upgrade involved
![platform upgrade involved flow](/img/ocp_platform_upgrade.png)
```

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,376 @@
...
 
 ## Requirements
 
-- OpenShift Container Platform (OCP) 4.19 or later
-- Consul 2.0.0 or later
+- A Consul deployment in Kubernetes.
+
+### Prerequisites
+
+- `kubectl` tool should be instaleld for non-OCP clusters and `oc` for OCP clusters.
```
aimeeu, line 29:
```suggestion
- `kubectl` tool is installed for non-OCP clusters and `oc` for OCP clusters.
```

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
aimeeu: Did you add the link to the rate limit reference doc in the original `consul-helm` repo's values.yaml file? According to L19 in reference/k8s/helm.mdx, lines 22-3033 are generated from the values.yaml file.

aimeeu (changes requested):
Suggestions to fix a broken image link and a typo. Also a question about a link added to helm.mdx

content/consul/v2.0.x/content/docs/reference/k8s/helm.mdx
Manishakumari-hc: @aimee removed the manually added link.
On generating a new helm.mdx using values.yaml locally there are few more changes getting added, which i would not like to introduce at the moment.

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -26,32 +26,42 @@ The Helm chart exposes several useful configurations and automatically sets up c
...
+- [`kubectl`](https://kubernetes.io/docs/reference/kubectl/) configured to point to the Kubernetes cluster you want to install Consul on.
+
+Depending on your requirements you can:
 
-1. Add the HashiCorp Helm Repository:
+- Perform a [default installation](#install-consul) by using the default Helm chart configuration. Use this option if you are testing Consul or setting up a test environment that does not have security requirements.
+- Perform a [custom installation](#custom-installation) by modifying Helm chart configuration parameters. This is the recommended approach if you are installing Consul in production.
+
+
+Add HashiCorp Helm Repository:
```
danielehc:
```suggestion
### Add HashiCorp Helm Repository

Before the installation, add HashiCorp repository to Helm.
```

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -117,17 +131,103 @@ NAME: consul
...
+| Non-OCP     | Platform manages CRDs, TCPRoute not required  | Standard API gateway | `global.installK8sNetworkingCRDs=false`, `global.crds.enableTcpRoute=false`                                                                                                        |
+| Non-OCP     | Platform manages CRDs, TCPRoute required      | Consul API gateway   | `global.installK8sNetworkingCRDs=false`, `global.crds.enableTcpRoute=false`, `global.crds.consulapi.enabled=true`                                                                  |
+
 ### Install Consul on OpenShift clusters
 
-[Red Hat OpenShift](https://www.redhat.com/en/technologies/cloud-computing/openshift) is a security-conscious, opinionated wrapper for Kubernetes. To install Consul on OpenShift-managed Kubernetes, set `global.openshift.enabled=true` in your values file:
+[Red Hat OpenShift](https://www.redhat.com/en/technologies/cloud-computing/openshift) is a security-conscious, opinionated wrapper for Kubernetes. To install Consul on OpenShift-managed Kubernetes, run the following commands to identify your OCP version and whether `TCPRoute` resources are present in your cluster:
+
```
danielehc, line 158:
```suggestion

Check OCP version.

```

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -117,17 +131,103 @@ NAME: consul
...
+```shell-session
+$ oc version
+```
+
```
danielehc, line 164:
```suggestion

Check if your cluster uses `TCPRoute`.

```

content/consul/v2.0.x/content/docs/deploy/server/k8s/helm.mdx
```diff
@@ -117,17 +131,103 @@ NAME: consul
...
+  </CodeBlockConfig>
+
+  </Tab>
+
+</Tabs>
+
+Refer to [`openshift` in the Helm chart reference](/consul/docs/reference/k8s/helm#v-global-openshift) for additional information.
+
+## Update your Consul on Kubernetes configuration
```
danielehc:
```suggestion
## Update your Consul configuration
```

We are in a dedicated K8s page so the "on Kubernetes" is implicit.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,376 @@
...
-- OpenShift Container Platform (OCP) 4.19 or later
-- Consul 2.0.0 or later
+- A Consul deployment in Kubernetes.
+
+### Prerequisites
+
+- `kubectl` tool is installed for non-OCP clusters and `oc` for OCP clusters.
+- Back up `Gateways`, `HTTPRoutes`, `TCPRoutes`, and `ReferenceGrants` in your cluster.
+  For example, run the following commands to create namespace-level, resource-specific backup files:
+
```
danielehc:
```suggestion
- `kubectl` tool is installed for non-OCP clusters and `oc` for OCP clusters.
- Back up `Gateways`, `HTTPRoutes`, `TCPRoutes`, and `ReferenceGrants` in your cluster.
 
 Use the following commands as a reference to create namespace-level, resource-specific backup files:

```

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,376 @@
...
+  Move to the `consul.hashicorp.com` API group to maintain compatibility with OCP 4.19 and later.
+
+- Non-OCP Kubernetes clusters that use `TCPRoute` and platform-managed Gateway CRDs.
+
+  Move to the `consul.hashicorp.com` API group to maintain compatibility.
+
+## Upgrade flow charts
+
+Use the following flow charts to choose the correct path for your upgrade.
```
danielehc, line 89:
```suggestion
```

Removing the extra H2 because the section can go inside the previous one and we do not crowd the right menu.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
```diff
@@ -1,78 +1,376 @@
...
+- Consul is deployed in Kubernetes and uses Helm-based configuration.
+- Gateway-related resources are backed up before the upgrade.
+
+
+The upgrade process involves these resource types:
+
+- `Gateway`
+- `HTTPRoute`
+- `TCPRoute`
+- `ReferenceGrant`
```
danielehc:
```suggestion
## OpenShift Container Platform (OCP) upgrade steps

The upgrade process 

To prepare for an OCP upgrade from 4.18 to 4.19+ you can follow these steps:

1. Make sure the OCP cluster s running OCP 4.18 or earlier and the target version is OCP 4.19 or later.
2. Verify Consul is deployed in the Cluster
3. Validate if the gateway CRDs under API group `gateway.networking.k8s.io` are pre-installed in the cluster. If pre-installed, set `global.installK8sNetworkingCRDs: false`.  
4. Make sure Gateway-related resources are backed up before the upgrade.

The backup and upgrade process involves these resource types:

- `Gateway`
- `HTTPRoute`
- `TCPRoute`
- `ReferenceGrant`

After all the preparations are complete, use the flowchart to identify the specific steps for your upgrade.

![Dual gateway new installation of Consul and OCP upgrade flow](/img/ocp-upgrade.png)

Finally, follow the steps in the section below according to your configuration.

```

The information was confusing and duplicated. I tried to condense all.

**One question:** Do we have to be on 4.18.x to perform these steps or the same upgrade steps apply to an earlier version of OCP? 

If being on 4.18 is a requirement we should include an extra step to upgrade OPS from <4.18 to 4.18.

danielehc: Also, do we have links for the OCP upgrade process we want to suggest?

danielehc (dismissed):
Made minor changes. The flow looks good now.

Please apply the requested changes and I think we are ready to merge.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/enable.mdx
```diff
@@ -1,8 +1,8 @@
 ---
```
danielehc, line 1: This file is also modified in https://github.com/hashicorp/web-unified-docs/pull/2500. The changes should not cause conflicts but be aware of it.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/peer.mdx
```diff
@@ -9,13 +9,13 @@ last_modified: 2026-03-31T20:45:24.000Z
```
danielehc, line 9: This file is also modified in https://github.com/hashicorp/web-unified-docs/pull/2500. The changes should not cause conflicts but be aware of it.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/listener.mdx
```diff
@@ -28,7 +28,7 @@ Configure Consul intentions to allow or prevent traffic between gateway listener
 ## Requirements
```
danielehc, line 28: This file is also modified in https://github.com/hashicorp/web-unified-docs/pull/2500. The changes should not cause conflicts but be aware of it.

Manishakumari-hc: Right .Lets merge 2500 first and then this PR.

content/consul/v2.0.x/content/docs/north-south/api-gateway/k8s/resource-upgrade.mdx
Manishakumari-hc: @danielehc applied the suggestion manually (faced error in auto application), 
and yes line 104 is telling about OCP version 4.18 or earlier ,which is right.

About your question of additional links for OCP upgrade process it will be the official link- `https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/pdf/updating_clusters/index#updating-cluster-prepare`

If you think this belongs in the documentation, your input on the right section to place it would be helpful.Feel free to do it if it is easier than reviewing again :)

aimeeu (approved):

mattschouten (approved):
Approving based on Aimee's approval and conversation with Daniele

PR #2843, opened 2026-07-14
Fix terraform-policy nesting

RubenSandwich:
Fixes a few problems in`terraform-policy`:
1. was nested inside of `terraform-docs-agents` which caused it not be found by `dev-portal`
2. Needed a redirect.jsonc file
3. We had an assumption because that at least one stable release will exist, while this is launching as a beta. So this also fixes our metadata generation in this case.

### DIFF
--- a/app/utils/allDocsPaths.ts
+++ b/app/utils/allDocsPaths.ts
@@ -29,13 +29,11 @@ export const getDocsPaths = async (
 				return []
 			}
 
-			const { version, releaseStage } = latestProductMetadata.value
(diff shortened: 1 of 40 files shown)


rmainwork (dismissed):

rmainwork (dismissed):

robin-norwood (dismissed):
lgtm!

robin-norwood (approved):
lgtm, thanks!

rmainwork (approved):

PR #2869, opened 2026-07-17
Fix release stage not being taken into account with algolia sync

RubenSandwich:
## Description

[Algolia sync is breaking due to `terraform-policy` having a non-stable release stage in it's latest version.](https://github.com/hashicorp/web-unified-docs/actions/runs/29584387669/job/87897608749#step:5:63) This PR fixes that.

## Testing

1. checkout this branch
2. `npm run prebuild -- --get-real-file-changed-metadata --build-algolia-index` should work locally


williamdalessandro (approved):
Worked when I ran it locally 👍

PR #2895, opened 2026-07-22
docs(nomad): correct coalesce empty-string behavior

### LINKED ISSUE #15599: hcl2 coalesce function returning empty string element when the second element is a non-empty string
### Nomad version
`Nomad v1.4.3 (f464aca721d222ae9c1f3df643b3c3aaa20e2da7)`

### Operating system and Environment details
MacbookPro Intel running macOS BigSur (11.7.1)

### Issue
Nomad 1.4x docs for the `colaesce` fuction states:

> `coalesce` takes any number of arguments and returns the first one that isn't null or an empty string.

But that is not the case, it will return the first argument that isn't null **even if it's an empty string**.

Checking the implementation of the `CoalesceFunc` [in go-cty](https://github.com/zclconf/go-cty/blob/36f1b4d0a3663ebca3accd71a4ac236104ad3992/cty/function/stdlib/general.go#L59-L93) looks like this might just be a documentation issue, as their the function implementation description matches the observed behavior:

> Returns the first of the given arguments that isn't null, or raises an error if there are no non-null arguments.

That said, it would be great if possible to have a `coalesce` function (or a different function) that allows coalescing not only `nil` values but also empty strings.

### Reproduction steps
```
coalesce("", "foo")
```
#### Expected Result
`"foo"`

#### Actual Result
` ""`

locker95:
## Summary

The `coalesce` docs said the function returns the first argument that isn't null **or an empty string**, and the examples showed `coalesce("", "b")` returning `b`.

That doesn't match the implementation. Nomad wires `coalesce` to go-cty's `stdlib.CoalesceFunc`, which only skips **null** values. An empty string is a real value, so:

```hcl
coalesce("", "foo")  # => ""
```

This came up in hashicorp/nomad#15599 and was confirmed as a docs issue.

This PR:

- Aligns the description with the null-only behavior
- Fixes the examples (`null` is skipped; `""` is not)
- Points people who want to skip empty strings at `compact`

Updated for the current docs version (`v2.0.x`) and `v1.11.x`.

Fixes hashicorp/nomad#15599

## Test plan

- [x] Compared against go-cty `CoalesceFunc` and Nomad's `jobspec2` function table
- [x] Examples match observed behavior (`coalesce("", "b")` => `""`)
- [ ] Docs preview / tech writer review

### DIFF
--- a/content/nomad/v1.11.x/content/docs/reference/hcl2/functions/collection/coalesce.mdx
+++ b/content/nomad/v1.11.x/content/docs/reference/hcl2/functions/collection/coalesce.mdx
@@ -3,7 +3,7 @@ layout: docs
 page_title: coalesce - Functions - Configuration Language
 description: |-
   The coalesce function takes any number of arguments and returns the
-  first one that isn't null nor empty.
+  first one that isn't null.
 # START AUTO GENERATED METADATA, DO NOT EDIT
 created_at: 2025-11-17T20:24:06-05:00
 last_modified: 2025-11-17T20:24:06-05:00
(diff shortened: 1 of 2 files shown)


tgross (approved):
LGTM

aimeeu (approved):
Thanks for the update!

PR #2897, opened 2026-07-22
docs(vault): correct TTL duration suffix docs for PKI/SSH roles

### LINKED ISSUE #31404: docs: create/update role allows days as a time suffix
**Describe the bug**
According to https://developer.hashicorp.com/vault/api-docs/secret/pki#create-update-role, 

> [ttl](https://developer.hashicorp.com/vault/api-docs/secret/pki#ttl-5) (string: "") - Specifies the Time To Live value to be used for the validity period of the requested certificate, provided as a string duration with time suffix. Hour is the largest suffix. The value specified is strictly used for future validity. If not set, uses the system default value or the value of max_ttl, whichever is shorter. See not_after as an alternative for setting an absolute end date (rather than a relative one).
> [max_ttl](https://developer.hashicorp.com/vault/api-docs/secret/pki#max_ttl) (string: "") - Specifies the maximum Time To Live provided as a string duration with time suffix. Hour is the largest suffix. If not set, defaults to the system maximum lease TTL.

However, empirically `d` (days) seems to work (which is convenient, because cert lifetimes on the order of 90 days are pretty common).

**Expected behavior**

As with e.g. [`not_before_duration`](https://developer.hashicorp.com/vault/api-docs/secret/pki#not_before_duration), say this accepts durations and link to https://developer.hashicorp.com/vault/docs/concepts/duration-format

locker95:
## Description

The PKI (and SSH) create/update role API docs claimed that `Hour` is the largest duration suffix for `ttl` / `max_ttl`. Empirically, Vault accepts day units (and the other standard duration units), which matches the [duration string format](https://developer.hashicorp.com/vault/docs/concepts/duration-format) used elsewhere (e.g. `not_before_duration`).

### Changes

- Replace the incorrect "Hour is the largest suffix" wording with a link to the duration format docs
- Apply the same fix across currently published Vault versions: `v2.x` (latest), `v1.21.x`–`v1.16.x`
- Covers PKI create/update role, related PKI ACME `max_ttl`, and SSH role `ttl`/`max_ttl` which used the same incorrect text

## Links

GitHub issue: Fixes hashicorp/vault#31404

## Test plan

- [x] Confirmed duration format docs list `d` (days) as a valid unit
- [x] Matched existing `not_before_duration` / `safety_buffer` duration-format link style
- [x] Verified no remaining "Hour is the largest suffix" phrases in published version trees

### DIFF
--- a/content/vault/v1.16.x/content/api-docs/secret/pki.mdx
+++ b/content/vault/v1.16.x/content/api-docs/secret/pki.mdx
@@ -3315,13 +3315,13 @@ request is denied.
 
 - `ttl` `(string: "")` - Specifies the Time To Live value to be used for the
   validity period of the requested certificate, provided as a string duration
-  with time suffix. Hour is the largest suffix. The value specified is strictly
+  using [duration format strings](/vault/docs/concepts/duration-format). The value specified is strictly
   used for future validity. If not set, uses the system default value or the
   value of `max_ttl`, whichever is shorter. See `not_after` as an alternative
   for setting an absolute end date (rather than a relative one).
 
 - `max_ttl` `(string: "")` - Specifies the maximum Time To Live provided as a
-  string duration with time suffix. Hour is the largest suffix. If not set,
+  string duration using [duration format strings](/vault/docs/concepts/duration-format). If not set,
   defaults to the system maximum lease TTL.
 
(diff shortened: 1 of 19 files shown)


crw:

PR #2941, opened 2026-07-27
Fixing broken image paths in inc-builds for unversioned products

williamdalessandro:
Images from unversioned products resolve to a broken URL in incremental builds. Unversioned products, (WAF, terraform-docs-common, validated-designs, hcp-docs) versionMetadata.version produce a double-slash in the asset path (e.g. "assets/validated-designs//img/...") when loaded in previews. Locally this issue doesnt occur, nor does it occur in production, only in the deployed preview incremental builds.

This is actively visible in this [preview link from another PR](https://unified-docs-frontend-preview-qvbhl2m9e-hashicorp.vercel.app/validated-designs/vault/operating-guides/adoption/0010-people-and-process#internal-developer-platform) with the image that fails to load that only shows the placeholder text "Internal developer platform". 

I recreated the bug in earlier commits to take these screenshots. The first pic isn't loading the image that should be coming from "CURRENT BUILD", but the one loaded from PRODUCTION works fine:
Screenshot 2026-07-27 at 10 22 58 AM

After the fix you can see the duplicated test image is loading from the "CURRENT BUILD":
Screenshot 2026-07-27 at 11 56 11 AM

If you want to recreate the bug:
- checkout a branch off main
- add a new image to an unversioned product doc that's supported by UDR (WAF, terraform-docs-common, hcp-docs)
- see that it loads locally totally fine
- push it up in a PR to create the preview
- Find that it fails to load


LeahMarieBush (approved):
LGTM!

PR #2945, opened 2026-07-27
Updating docs for general diagnostics command

lakshyaa-nathan-15:
Added docs for the new general diagnostics command. 

The generalized diagnostics command can retrieve configuration diagnostics (via configuration ID), or deployment step diagnostics (via step ID). This unified command improves usability and enables automation agents to retrieve structured failure details without needing API calls.

### Terraform


content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
```diff
@@ -0,0 +1,102 @@
...
+- Deployment step IDs use the prefix `sds-`.
+
+The command returns an empty result when no diagnostics are present.
+
+## Options
+
+- `-id`: The ID of the stack configuration or deployment step to retrieve diagnostics for.
+    - Required.
+    - String.
+    - Stack configuration IDs use the prefix `stc-`. Deployment step IDs use the prefix `sds-`.
```
jondavidjohn: Oh! Did I miss that we already support both? I thought we only did config diagnostics currently.

lakshyaa-nathan-15: Hi! I believe release 1.3.0 supports both.

jondavidjohn: Nice!

jondavidjohn (dismissed):
This looks good to me 👍

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
```diff
@@ -0,0 +1,102 @@
...
+description: The terraform stacks diagnostics command retrieves diagnostics for a stack configuration or deployment step.
+# START AUTO GENERATED METADATA, DO NOT EDIT
+created_at: 2026-06-16T00:00:00Z
+last_modified: 2026-06-16T00:00:00Z
+# END AUTO GENERATED METADATA
+---
+
+# `terraform stacks diagnostics` reference
+
+Use the `terraform stacks diagnostics` command to retrieve diagnostics for a stack configuration or deployment step.
```
mattschouten:
```suggestion
Use the `terraform stacks diagnostics` command to retrieve warnings and errors for a Stack configuration or deployment step.
```

I'm going to make similar suggestions throughout the page.  The reason is that "diagnostics" here seems to mean exactly "warnings and errors", so the word "diagnostics" doesn't add much.  The user is already in the context of the diagnostics page, so it is not likely to confuse them.

If this take is off-base, please say so.

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Usage
+
+```shell-session
+$ terraform stacks <global-stacks-flags> diagnostics -id <value> <options>
+```
+
+## Description
+
+The `terraform stacks diagnostics` command retrieves diagnostic information — warnings and errors — for either a stack configuration or a deployment step. Pass the ID of the resource you want to inspect using the `-id` flag:
```
mattschouten:
```suggestion
The `terraform stacks diagnostics` command retrieves warnings and errors for either a Stack configuration or a deployment step.
```

The details about the ID and the prefix can be covered in the Options section.

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+- Stack configuration IDs use the prefix `stc-`.
+- Deployment step IDs use the prefix `sds-`.
+
+The command returns an empty result when no diagnostics are present.
```
mattschouten:
```suggestion
The command returns an empty result when no errors or warnings are present.
```

I'm not quite clear on what "are present" means here.  Does that mean for the last run, ever, ...?

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
```diff
@@ -0,0 +1,102 @@
...
+```shell-session
+$ terraform stacks <global-stacks-flags> diagnostics -id <value> <options>
+```
+
+## Description
+
+The `terraform stacks diagnostics` command retrieves diagnostic information — warnings and errors — for either a stack configuration or a deployment step. Pass the ID of the resource you want to inspect using the `-id` flag:
+
+- Stack configuration IDs use the prefix `stc-`.
+- Deployment step IDs use the prefix `sds-`.
```
mattschouten:
```suggestion
```

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
mattschouten:
```suggestion
    - Prefix the ID with `stc-` for a Stack configuration ID. Prefix the ID with `sds-` for a deployment step ID.
```

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
mattschouten: I _think_ a Stack configuration is a specific version within a set of versions.  The thing that would show up in the UI under "Recent Versions" as "#53", and have the end of the URL look like `configurations/stc-FZ2uHfN5QM5vUZq5`.

I think a deployment step is when I click into the configuration, then the deployment (e.g., "dev"), then get the ID by hitting the down arrow for "Inspect run data", and the ID looks like `sds-yUQyF49PxmMATR26`.

Will readers of this page know that already?  If not, is there an easy way to explain that context to them on this page?

Or should we link to something like https://developer.hashicorp.com/terraform/cloud-docs/api-docs/stacks/deployments#list-stack-deployment-steps and https://developer.hashicorp.com/terraform/cloud-docs/api-docs/stacks/configurations#list-stack-configurations?

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Output
+
+### Human-readable format
+
+Warnings and errors are printed using the standard Terraform diagnostic format. Warnings are prefixed with `│ Warning:` and errors are prefixed with `│ Error:`, each followed by a summary, detail, and optional source location.
+
+### JSON format
+
+The JSON output uses the following schema:
```
mattschouten: This is closer to an example than a schema.  Three options I see:
1. Make it really a schema
2. Make it two examples, showing configuration/error and deployment step/warning to cover the changes
3. Omit entirely, because anyone using this would try it out and see the shape.

Which seems best?

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
```diff
@@ -0,0 +1,102 @@
...
+      "snippet": {
+        "code": "  target_id = each.value.target_id",
+        "context": "resource \"aws_lb_target_group_attachment\" \"this\""
+      }
+    }
+  ]
+}
+```
+
+The `range` and `snippet` fields are optional and may be omitted when source location information is unavailable.
```
mattschouten:
```suggestion
The `range` and `snippet` fields are omitted when source location information is unavailable.
```

Assuming they're always omitted if source location isn't available?

content/terraform/v1.15.x/docs/cli/commands/stacks/index.mdx
```diff
@@ -27,6 +27,7 @@ To use CLI commands with Stacks on Terraform Enterprise, set the `TF_STACKS_HOST
 - [`validate`](/terraform/cli/commands/stacks/validate) - Checks whether the local component and deployment configurations are valid.
 - [`create`](/terraform/cli/commands/stacks/create) - Create a Stack in HCP Terraform.
 - [`list`](/terraform/cli/commands/stacks/list) - List Stacks for a given organization or project.
+- [`diagnostics`](/terraform/cli/commands/stacks/diagnostics) - Retrieve diagnostics for a stack configuration or deployment step.
```
mattschouten:
```suggestion
- [`diagnostics`](/terraform/cli/commands/stacks/diagnostics) - Retrieve warnings and errors for a stack configuration or deployment step.
```

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
```diff
@@ -0,0 +1,102 @@
+---
```
mattschouten, line 1: Please also add this new page to the nav in content/terraform/v1.15.x/data/cli-nav-data.json.

mattschouten (changes requested):
This is pretty good, overall.  Feedback falls into a few buckets:
- Being specific about what "diagnostics" means, and omitting extra words
- Clarifications
- Minor cleanup

Feel free to message me if you have questions!

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
lakshyaa-nathan-15: Hi! I replaced this line with "The command returns an empty result when no errors or warnings are associated
with the specified configuration version or deployment step."

Let me know if this is more clear.

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
lakshyaa-nathan-15: We could definitely clarify the context of both on the page. Perhaps a couple sentences each for the configuration ID and step ID will help. Here is what I'm thinking for the context:

> A stack configuration ID (`stc-`) identifies a specific uploaded version of your stack configuration. You receive this ID as output from the `terraform stacks configuration upload` or `configuration fetch` commands.
> 
> A deployment step ID (`sds-`) identifies a single plan or apply step within a deployment run. You can find step IDs in the output of the `terraform stacks deployment-run show` command.

Let me know if you agree, and I will add this.

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
lakshyaa-nathan-15: I think it would be best to omit the whole "Output" section entirely. The format seems pretty self-explanatory and the section is inconsistent with other doc pages.

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
mattschouten: That's much clearer to someone that doesn't already know the details of the command.  Very nice!

I'm going to make a suggestion that might be a step too basic (too spelling-it-all-out), just as food for thought.  "associated with" is not very specific.
- Configuration version would have no warnings or errors if no warnings or errors were generated when reading and parsing the version.  (There may be more possibilities.)
- Deployment step would have no warnings or errors if no errors or warnings were generated when running the step.

So you _could_ write something like:

The command returns an empty result when no errors or warnings are generated when loading and parsing the configuration version or running that deployment step.

I don't think going to that level of specificity is necessary here.

The reason for being more specific is to avoid questions in the reader's mind.  Some examples:
- If I have a completely valid configuration version, but there's an error when it runs (say, AWS permissions), is there an error associated with the configuration version?
- If I have a deployment step (B) that doesn't work because of a prior deployment step (A), is there an error associated with the prior step (A)?

I think in both cases the answer would be "no".

(It feels a little lawyer-ish, or tester-edge-case-finding-ish.  But whether writing code or docs, I've found that looking for those gaps helps make the final product more robust.  And you can always decide "nope, that's too detailed/clunky.")

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
mattschouten: I agree.  I like the way you described those, and using the CLI commands rather than what the UI shows is a good idea.

Minor nitpick:  please capitalize `Stack` when it's not a command example.  It's a product name.

(Usually the nitpick runs the other way, where people want to capitalize concepts that aren't product names!)

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
mattschouten: Works for me!

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
lakshyaa-nathan-15: Makes sense, thanks for pointing that out!

content/terraform/v1.15.x/docs/cli/commands/stacks/diagnostics.mdx
lakshyaa-nathan-15: Thanks for the thoughtful feedback! I went with "exist for" instead of "associated with" to tighten it up. I think it reads more naturally while still being more precise than the original. 

The specificity seems somewhat out of scope for this command description. I can imagine someone running `terraform stacks diagnostics` just needs to know what an empty result means, not the precise conditions under which errors are generated at each stage. I'm happy to revisit if you feel the edge cases still need explicit coverage!

mattschouten (approved):
Looks great, nice work!

PR #2974, opened 2026-07-30
terraform: Document the `planned_action_invocation` JSON log entry type that was added in v1.14

SarahFrench:
Reference: https://github.com/hashicorp/terraform/pull/37467
Reference: https://github.com/hashicorp/terraform/pull/38936

The `planned_action_invocation` type of JSON log is produced by the Actions feature that was released in v1.14. This PR documents that log entry type for v1.14 and recent versions.

I used this tutorial (https://developer.hashicorp.com/terraform/tutorials/configuration-language/actions) to perform actions with the -json flag, so the examples are copied as-is from those manual experiments.


SarahFrench:
> ## Broken Link Checker
> **This PR contains broken links, but won't be blocked.** Use this report to improve content quality:
> 
> ...

Consul isn't relevant here

radeksimko (approved):

PR #2985, opened 2026-07-31
fix: update copyright headers

LeahMarieBush:
### Description

This PR fixes the copyright headers on these files to fix the Check copyright Headers error that has been failing. Also adds .tf files to the github action to prevent this from happening in the future. It looks like it did not run on the original PR since `.tf` files were not included in the GHA

### Testing

Make sure the Check Copyright Headers check passes on this PR: https://github.com/hashicorp/web-unified-docs/actions/runs/30657871963/job/91246668421?pr=2985

### DIFF
--- a/.github/workflows/copyright-headers-check.yml
+++ b/.github/workflows/copyright-headers-check.yml
@@ -11,6 +11,7 @@ on:
       - '**.ts'
       - '**.tsx'
       - '**.mjs'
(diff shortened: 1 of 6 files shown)


ryan-yu-ibm (approved):
Makes sense to me! Needed to have copyright added to top of .tf files and mark it as a file type to check in the GHA.

PR #2991, opened 2026-08-03
docs(vault): correct optimistic failure tolerance figures

### LINKED ISSUE #2914: Incorrect optimistic failure tolerance figures
### Product

Vault

### Issue type

Other issue

In the documentation for Vault, the "Optimistic failure tolerance" (OFT) figures cited are incorrect.

For example, in a cluster with **3 redundancy zones and 2 nodes in each zone**, the quorum size is 2, so the OFT **_should be 4, instead of 3_**:
https://developer.hashicorp.com/vault/docs/internals/integrated-storage#redundancy-zones
https://developer.hashicorp.com/vault/docs/enterprise/redundancy-zones#optimistic-failure-tolerance
The statements saying that "The cluster then maintains 3 voting nodes with 2 remaining standbys. The cluster can handle an **_additional 2 gradual failures_** before it loses quorum." are thus inaccurate as well. It should have been "an additional 3 gradual failures".

_(eg. The correct OFT figure of 4 is cited here on a different page:
https://developer.hashicorp.com/vault/tutorials/raft/raft-redundancy-zones#:~:text=Optimistic%20Failure%20Tolerance%3A%20%20%20%204)_

Similar inaccuracies apply to the other figures cited in the table for clusters with **3 redundancy zones and 3 nodes per zone**, as well as clusters with **5 redundancy zones and 2 nodes per zone**.

locker95:
The redundancy-zones OFT table and the 3-zone/2-node walkthrough were a bit low (e.g. 3 instead of 4 for 3×2).

Updated the optimistic failure tolerance column and the example prose to match gradual promote-then-fail counting (standbys + voters − quorum). Same numbers as the raft redundancy-zones tutorial.

Fixes #2914

### DIFF
--- a/content/vault/v1.17.x/content/docs/internals/integrated-storage.mdx
+++ b/content/vault/v1.17.x/content/docs/internals/integrated-storage.mdx
@@ -324,9 +324,9 @@ on how many redundancy zones and servers per redundancy zone that you choose.
 Redundancy zones | Servers per zone | Quorum size | Failure tolerance | Optimistic failure tolerance
 :--------------: | :--------------: | :---------: | :---------------: | :--------------------------:
 2                | 2                | 2           | 0                 | 2
-3                | 2                | 2           | 1                 | 3
-3                | 3                | 2           | 1                 | 5
-5                | 2                | 3           | 2                 | 6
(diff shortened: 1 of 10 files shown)


crw:

PR #2992, opened 2026-08-03
docs(cli): document stale saved plan rejection on apply

### LINKED ISSUE #2827: Document that Terraform detects stale saved plan files during apply
### Product

Terraform

### Issue type

New content

In the Terraform CLI documentation for saved plan workflows, we should document that Terraform detects stale saved plan files during apply and refuses to apply them.

Relevant pages:

- https://developer.hashicorp.com/terraform/cli/commands/plan#out-filename
- https://developer.hashicorp.com/terraform/cli/commands/apply#saved-plan-mode

Use case:

As an engineer building automation around:

1. `terraform plan -out=plan_artifact`
2. `terraform apply plan_artifact`

I want to know whether I can rely on Terraform to detect that the saved plan was created against an older state snapshot, or whether my automation needs to implement its own stale-plan detection before calling `terraform apply`.

Current Terraform behavior:

Terraform reads the prior state snapshot from the plan file and compares its state metadata with the current state metadata. If the saved state serial differs from the current state serial, it emits `Saved plan is stale`.
`Saved plan is stale`

Source code reference:

https://github.com/hashicorp/terraform/blob/main/internal/backend/local/backend_local.go

Suggested documentation change:

Please document this behavior in the saved plan sections of the Terraform CLI docs, especially `terraform apply` saved plan mode and/or `terraform plan -out=FILENAME`.

As a user, especially when using remote backends where multiple developers could be pushing state, it is important to have clarity that Terraform will handle state staleness detection. 

Reproduce:

Image

In the Terraform sandbox
```bash
terraform init
terraform plan -out=tfplan
cat >> main.tf <<'EOF'

resource "terraform_data" "advance_state" {
  input = "make the saved plan stale"
}
EOF
terraform plan -out=updated_tfplan
terraform apply updated_tfplan
terraform apply tfplan
```
```bash
Error: Saved plan is stale
│ 
│ The given plan file can no longer be applied because the state was changed by another operation after the plan was created.
```

locker95:
Saved-plan apply already fails with **Saved plan is stale** / lineage mismatch when state moved on, but the CLI pages didn't say that.

Added a short note on `apply` (and a matching line under `plan -out`) so people know to re-plan instead of retrying the same file.

Fixes #2827

### DIFF
--- a/content/terraform/v1.13.x/docs/cli/commands/apply.mdx
+++ b/content/terraform/v1.13.x/docs/cli/commands/apply.mdx
@@ -34,6 +34,8 @@ When you pass a [saved plan file](/terraform/cli/commands/plan#out-filename) to 
 
 Use [`terraform show`](/terraform/cli/commands/show) to inspect a saved plan file before applying it.
 
+Terraform checks that the saved plan still matches the current state before applying it. If another operation changed the state after the plan was created (for example another apply bumped the state serial), Terraform refuses to apply the file and returns **Saved plan is stale**. If the plan was created from a different state lineage, Terraform returns **Saved plan does not match the given state**. In either case, create a new plan with `terraform plan -out=...` and apply that file instead.
+
 When using a saved plan, you cannot specify any additional planning modes or options. These options only affect Terraform's decisions about which
 actions to take, and the plan file contains the final results of those
 decisions.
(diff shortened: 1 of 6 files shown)


crw:

PR #2997, opened 2026-08-04
docs(raft): add -dr-token note to list-peers and remove-peer commands

ShamanthGaddam:
## Summary

Add a reusable partial documenting the `-dr-token` flag for `vault operator raft` commands on DR secondary clusters, and include it in both the `list-peers` and `remove-peer` sections.

## Changes

- **New file:** `content/vault/v1.20.x/content/partials/raft-dr-token.mdx`
  A reusable `<Note>` callout explaining:
  - Use `-dr-token <dr-operation-token>` to run commands on a DR secondary cluster
  - DR secondaries mirror their primaries — changes should be made on the primary first
  - Links to the DR operation token documentation

- **Updated:** `content/vault/v1.20.x/content/docs/commands/operator/raft.mdx`
  Added `@include 'raft-dr-token.mdx'` to both the `list-peers` and `remove-peer` sections

## Related

Resolves VAULT-4408


content/vault/v1.20.x/content/docs/commands/operator/raft.mdx
```diff
@@ -103,6 +103,8 @@ Usage: vault operator raft list-peers
 	  $ vault operator raft list-peers
 ```
```
sgsource, line 105: We should add a "parameters" h3 here to keep it consistent with the join command above.

content/vault/v1.20.x/content/docs/commands/operator/raft.mdx
```diff
@@ -103,6 +103,8 @@ Usage: vault operator raft list-peers
...
 
+@include 'raft-dr-token.mdx'
```
sgsource, line 106: This is more of a nitpick, but I think we should put this partial in a more specific dir instead of the top level of partials. Something like `cli/...`.

content/vault/v1.20.x/content/partials/raft-dr-token.mdx
```diff
@@ -0,0 +1,14 @@
+<Note title="DR secondary clusters">
```
sgsource, line 1: I think it's better we make this a linkable tag.

sgsource (commented):
Left some comments

content/vault/v1.20.x/content/partials/raft-dr-token.mdx
ShamanthGaddam: Closing this PR since @sgsource  and I are working on the same ticket. We decided to move forward with this PR instead: https://github.com/hashicorp/web-unified-docs/pull/2999

PR #3004, opened 2026-08-05
'-start' doesn't work when RAFT is the destination storage.

AkashGuptaHC:
Here is the bug ID [https://hashicorp.atlassian.net/browse/VAULT-48858] to support this evidence.

Slack Thread: https://ibm-hashicorp.slack.com/archives/C09L31FC7CL/p1785399602385509

This update should go to all the branches of this page because this bug is applicable to all the Vault versions available as of today.

### DIFF
--- a/content/vault/v2.x/content/docs/commands/operator/migrate.mdx
+++ b/content/vault/v2.x/content/docs/commands/operator/migrate.mdx
@@ -131,7 +131,7 @@ The following flags are available for the `operator migrate` command.
 
 - `-config` `(string: <required>)` - Path to the migration configuration file.
 
-- `-start` `(string: "")` - Migration starting key prefix. Only keys at or after this value will be copied.
+- `-start` `(string: "")` - Migration starting key prefix. Only keys at or after this value will be copied. Please note, this flag isn't applicable when Integrated Storage (RAFT) is the destination storage.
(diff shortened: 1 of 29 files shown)


nickfcodes:
This was fixed and merged. This PR to update the docs can be closed.  https://github.com/hashicorp/vault-enterprise/pull/17265

content/vault/v2.x/content/docs/commands/operator/migrate.mdx
```diff
@@ -131,7 +131,7 @@ The following flags are available for the `operator migrate` command.
 
 - `-config` `(string: <required>)` - Path to the migration configuration file.
 
-- `-start` `(string: "")` - Migration starting key prefix. Only keys at or after this value will be copied.
+- `-start` `(string: "")` - Migration starting key prefix. Only keys at or after this value will be copied. Please note, this flag isn't applicable when Integrated Storage (RAFT) is the destination storage.
```
schavis, line 134:
```suggestion
- `-start` `(string: "")` - Migration starting key prefix. Vault only copies keys
  at or after the start value for deployments that do not use integrated storage
  (RAFT).
```

schavis (changes requested):

PR #3012, opened 2026-08-05
Explicitly provide instana agent key to GH build

rmainwork:
Currently build metrics are not being sent to Instana. Since dev portal's build happens in vercel (not github actions) it has access to all environment variables (even those marked as sensitive). According to [vercel's documentation](https://vercel.com/docs/environment-variables/sensitive-environment-variables) - _"Sensitive environment variables are [environment variables](https://vercel.com/docs/environment-variables)
whose values are non-readable once created. They help protect sensitive information stored in environment variables, such as API keys."_. I suspect that the reason that these metrics aren't showing up in Instana is that (although we do run a `vercel env pull`) in the build step, they are replaced with `[SENSITIVE]` instead of the actual value causing the build metrics to be not emitted to Instana correctly.

image

Additionally, it makes sense to rename the environment variable for this value slightly to something more meaningful as `INSTANA_OTLP_AGENT_TOKEN` implies that this secrets is only used for emitting OTel metrics (which is not correct).

### DIFF
--- a/.github/workflows/deploy-udr.yml
+++ b/.github/workflows/deploy-udr.yml
@@ -46,7 +46,7 @@ jobs:
         run: vercel build --prod --token=${{ secrets.VERCEL_TOKEN }}
         env:
           INSTANA_OTLP_ENDPOINT: ${{ secrets.INSTANA_OTLP_ENDPOINT }}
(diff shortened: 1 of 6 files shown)


RubenSandwich (approved):
👍🏻

ryan-yu-ibm (approved):

PR #3026, opened 2026-08-07
docs(terraform): clarify when moved blocks can be removed

### LINKED ISSUE #728: information in 'Removing moved Blocks' seems to be false
### Terraform Version

```shell
terraform v1.9.4
on windows amd_64

(We use TFE so its not my local system and this happens on every tf core version we have in use  1.3.x - 1.5.X)
```

### Affected Pages

https://developer.hashicorp.com/terraform/language/modules/develop/refactoring

### What is the docs issue?

The Chapter "Removing moved Blocks" states "Removing a moved block is a generally breaking change because any configurations that refer to the old address will plan to delete that existing object instead of move it. We strongly recommend that you retain all historical moved blocks from earlier versions of your modules to preserve the upgrade path for users of any previous version." this wording is a bit confusing (We understand this that the moved blocks can never be removed) our interpretation does not seem to be true as after an apply (using moved in modules is a different story as every code using that module needs to be applied first before removing it) with the moved blocks we can safely remove the moved block as the state has been re-written.

### Proposal

I'm not sure if this is an TFE only thing but IMO the docs should state something along the lines of this block can be removed after an apply has been done otherwise one would have a history if like 10 moved blocks as time goes on ?

Maybe something like this 

"Removing a moved block can generally be considered a breaking change because configurations referencing the old address might plan to delete the existing object rather than move it. However, once the moved block has been applied and the state has been updated, it is safe to remove the block, as the state has been rewritten. We strongly recommend retaining historical moved blocks in earlier versions of your modules to ensure a smooth upgrade path for users of any previous version."

### References

_No response_

locker95:
The old wording read like moved blocks could never go away. After apply the state already has the new address, so private modules can drop them once every consumer has applied — public modules still shouldn't.

Tightened that section on the current Terraform doc versions.

Fixes #728

### DIFF
--- a/content/terraform/v1.12.x/docs/language/modules/develop/refactoring.mdx
+++ b/content/terraform/v1.12.x/docs/language/modules/develop/refactoring.mdx
@@ -113,7 +113,7 @@ resource (a `data` block).
 
 ## Create multiple instances
 
-You can use a `moved` block when creating multiple instances of a resource or module call while preserving the original instance.
+You can use a `moved` block when creating mulitple instances of a resource or module call while preserving the original instance.
 
 ### Enable `count` or `for_each` for a resource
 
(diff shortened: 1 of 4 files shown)


crw:

PR #3027, opened 2026-08-07
docs(vault): make mount_accessor optional for entity-alias creation

elliesterner:
Prior to 2.0.0, entity alias creation required the mount_accessor and name to be provided. However, with the agentic workflow work (oauth resource servers), operators can omit the mount_accessor if they provide the issuer_id and external_id. Vault generates a mount_accessor for the alias under the hood. name is still required for agentic and non-agentic aliases. 

### Terraform


content/vault/v2.x/content/api-docs/secret/identity/entity-alias.mdx
```diff
@@ -42,7 +42,7 @@ This endpoint creates a new alias for an entity.
 
 - `canonical_id` `(string: <required>)` - Entity ID to which this alias belongs to.
 
-- `mount_accessor` `(string: <required>)` - Accessor of the mount to which the
+- `mount_accessor` `(string: <optional>)` - Accessor of the mount to which the
```
schavis, line 45:
```suggestion
- `mount_accessor` `(string:"")` - Accessor of the mount to which the
```
For option parameters, we use the actual default value, the empty string, or `<unset>` (if the value is actually `null` rather than empty.

schavis (changes requested):

content/vault/v2.x/content/api-docs/secret/identity/entity-alias.mdx
davidadeleon: This is an interesting one as its optional if and only if an issuer and external_id are provided. The way this reads to me is that its generally optional. This would break alias creation attempts for any requests that don't satisfy those requirements.

biazmoreira: mount_accessor should remain marked as required, since requests can fail when it is omitted, maybe? I prefer leaving it required. The description can note that it is not required when both issuer and external_id are provided.

VioletHynes: I think let's leave it as required and leave the description that it's not required in some cases yeah.

VioletHynes: I think this makes sense to close actually, since we already say that it's optional:
> You can omit `mount_accessor` during creation when you set `issuer`

@biazmoreira @davidadeleon does that make sense to you?

biazmoreira: Yes

VioletHynes:
Closing this as per discussion in thread. We already make it clear this is optional

PR #3034, opened 2026-08-07
docs(terraform): clarify create_before_destroy with count/for_each

### LINKED ISSUE #720: `create_before_destroy` behavior on `count`/`for_each` resources is not clearly documented
### Terraform Version

```shell
Terraform v1.4.2
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v4.67.0

is 1.11.3. You can update by downloading from https://www.terraform.io/downloads.html
```

### Affected Pages

* https://developer.hashicorp.com/terraform/language/meta-arguments/count
* https://developer.hashicorp.com/terraform/language/meta-arguments/for_each
* https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle

### What is the docs issue?

The behavior of `create_before_destroy` on a resource collection type using `count` or `for_each` isn't clear from documentation. There are two ways it could be interpreted to behave:

1. Every resource in the collection is an individual resource and `create_before_destroy` is applied to each separately. If a plan wants to both add and delete resources with separate `for_each` keys or `count` indices from the collection, the creates and destroys happen in an undefined order.
2. `create_before_destroy` applies to the collection as a whole. If a plan wants to both add and delete resources with separate `for_each` keys or `count` indices from the collection, all resources to add will be created before any resources to remove are destroyed.

We had a brief production outage because we assumed the latter, but we think we observed the former. The docs do not make it clear which behavior to expect.

### Proposal

Explicitly document how `create_before_destroy` and maybe `lifecycle` arguments in general are applied to resources in a collection and whether any of them apply to the collection as a whole.

### References

_No response_

locker95:
When a resource uses `count` or `for_each`, lifecycle rules like `create_before_destroy` apply per expanded instance, not to the collection as a whole. That distinction wasn't spelled out and is easy to get wrong in production.

Adds a short section under `create_before_destroy` on v1.12–v1.15.

Fixes #720

### DIFF
--- a/content/terraform/v1.12.x/docs/language/meta-arguments/lifecycle.mdx
+++ b/content/terraform/v1.12.x/docs/language/meta-arguments/lifecycle.mdx
@@ -46,6 +46,14 @@ to `false` on resource `B` because that would imply dependency cycles in the gra
 
 When the resource contains a provisioner that runs during the `destroy` operation,  setting `create_before_destroy` to `true` also prevents the provisioner from running. 
 
+#### `create_before_destroy` with `count` and `for_each`
+
+When a resource uses [`count`](/terraform/language/meta-arguments/count) or [`for_each`](/terraform/language/meta-arguments/for_each), Terraform expands it into multiple independent instances. Lifecycle rules such as `create_before_destroy` apply to each instance separately, not to the collection as a whole.
+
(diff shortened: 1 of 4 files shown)


crw:

PR #3035, opened 2026-08-07
docs(vault): clarify Terraform secrets engine anchor token scope

### LINKED ISSUE #1267: Update HCP Terraform Secrets Engine docs to clarify API token types, scope limitations, and usage examples
Current Terraform Secrets Engine documentation [here](https://developer.hashicorp.com/vault/docs/secrets/terraform) is misleading:

- The documentation recommends using a user API token as the root/anchor credential, but does not call out the limitation that user tokens cannot create user tokens for other users. This could lead to a misunderstanding that the Terraform Secrets engine can create short lived user API tokens
- The example shows a user API token for the root/anchor credential and creation of a user role, but does not call out that this would only work for the root/anchor user

locker95:
Makes it clearer that a user API token used as the plugin config credential can only create tokens for that same user—not other users, teams, or the org. Team/org tokens are the right anchor for broader automation.

Touches the HCP Terraform secrets engine page on the current vault docs versions.

Fixes #1267

### DIFF
--- a/content/vault/v1.21.x/content/docs/secrets/terraform.mdx
+++ b/content/vault/v1.21.x/content/docs/secrets/terraform.mdx
@@ -63,6 +63,13 @@ functions. An operator or configuration management tool usually completes these 
     [Organization, legacy team, team, and user roles](#organization-legacy-team-team-and-user-roles) 
     section for additional considerations.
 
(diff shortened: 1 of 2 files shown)


crw:

PR #3039, opened 2026-08-08
docs(terraform): drop jsondecode from mocking override examples

### LINKED ISSUE #732: Remove jsondecode function used in mocks examples
### Terraform Version

```shell
v1.7.0+
```

### Affected Pages

https://developer.hashicorp.com/terraform/language/tests/mocking#overrides-syntax

### What is the docs issue?

In the examples provided to support the explanation, you can find

```
# ./modules/s3_data/main.tf
variable "data_bucket_name" {
  type = string
}
data "aws_s3_object" "data_bucket" {
  bucket = var.data_bucket_name
  key    = "credentials.json"
}
output "data" {
  value = jsondecode(data.aws_s3_object.data_bucket.body)
}
```

Once you run `terraform test` you get an error in some of the previous tests defined something like

> data.aws_s3_object.data_bucket.body is ....Call to function "jsondecode" failed: extraneous data after JSON object."

because 'data.aws_s3_object.data_bucket.body' is already a JSON string -> unneeded `jsondecode`

### Proposal

I would suggest to remove `jsondecode`

```
variable "data_bucket_name" {
  type = string
}
data "aws_s3_object" "data_bucket" {
  bucket = var.data_bucket_name
  key    = "credentials.json"
}
output "data" {
  value = data.aws_s3_object.data_bucket.body
}
```

### References

_No response_

locker95:
The override examples decoded the S3 body in the module and re-encoded it for local_file. That made the override_module sample awkward (HCL object vs string) and can fail jsondecode depending on the mocked body. Passing the body through as a string keeps the shapes consistent.

v1.12.x–v1.15.x.

Fixes #732

### DIFF
--- a/content/terraform/v1.12.x/docs/language/tests/mocking.mdx
+++ b/content/terraform/v1.12.x/docs/language/tests/mocking.mdx
@@ -210,7 +210,7 @@ module "credentials" {
 
 resource "local_file" "credentials_json" {
   filename = "credentials.json"
-  content  = jsonencode(module.credentials.data)
+  content  = module.credentials.data
 }
 ```
 
@@ -227,7 +227,7 @@ data "aws_s3_object" "data_bucket" {
 }
 
 output "data" {
-  value = jsondecode(data.aws_s3_object.data_bucket.body)
+  value = data.aws_s3_object.data_bucket.body
 }
 ```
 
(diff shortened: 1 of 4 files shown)


crw:

PR #3043, opened 2026-08-08
docs(sdkv2): Planned Value mismatch is not only Optional+Computed

### LINKED ISSUE #726: "Planned Value does not match Config Value" incorrectly states conditions when the error will occur
Per https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/data-consistency-errors#planned-value-does-not-match-config-value:

> This occurs for attribute schema definitions that are Optional: true and Computed: true

However, it also occurs for Required fields, such as https://github.com/hashicorp/terraform/issues/37431

locker95:
The data consistency page said Planned Value vs Config Value only happens for Optional+Computed attributes. Same class of mismatch can show up for Required fields when the provider rewrites the value. Broadened the wording.

Fixes #726


crw:

PR #3044, opened 2026-08-08
docs(terraform): how to default optional object attributes

### LINKED ISSUE #727: make detailed type constraints easier to discover from the input variables docs
### Terraform Version

```shell
Terraform v1.9.3
on darwin_amd64
```

### Use Cases

I try to define an input variable defined as an object, where I would like to set optional attributes to a default value.

Ideal declaration:

```hcl
variable "db" {
  description = "Database definitions"
  type        = object({
    name = string
    user_name = string
    deletion_protection = optional(boolean)
    backup = optional(boolean)
  })
  default = {
    deletion_protection = true
    backup = true
  }
}

```

In the case above I would like deletion_protection to default to true. Terraform reports:

`This default value is not compatible with the variable's type constraint: attributes "name" and "user_name" are required.`

### Workaround Removing defaults

allow terraform to run, but now there is no documentation describing the default since it is handled elsewhere.

```hcl
variable "db" {
  description = "Database definitions"
  type        = object({
    name = string
    user_name = string
    deletion_protection = optional(boolean)
  })
}
```

**Attempt to workaround 1 - unsuccessful**

assign the default by using a local variable.

```
locals {
  db = merge( {deletion_protection = true}, var.db)
}
```
Now, if specifying only the required attributes, the optional key `deletion_protection` is still `null`, because `deletion_protection` is assigned by terraform as `bool(null)`.

**Work around 2 - successful** 

Using a loop to assign the default:

```hcl
locals {
  db = { for key, value in var.db :
    key => (
      (key == "deletion_protection" || key == "backup") && value == null ? true : value
      )
  }
}
```

**Work around 3 - successful** 

Trying to make the code more generic and efficient when default values 

```
locals {
  default_db_values = {
    deletion_protection = true
    backup = false
  }
  
  db = { for key, value in var.db :
    key => (
      value == null && contains(keys(local.default_db_values), key) ? local.default_db_values[key] : value
    )
  }
}
```

### Workaround Assigning required attributes default value of null

```hcl
variable "db" {
  description = "Database definitions"
  type        = object({
    name = string
    user_name = string
    deletion_protection = optional(bool)
    backup = optional(bool)
  })
  default = {
    name = null
    user_name = null
    deletion_protection = true
    backup = true
  }
  validation {
    condition = var.db.name != null && var.db.user_name != null
    error_message = "Both database name and user_name must be specified"
  }
}

``` 

In the scenario above the mandatory keys are assigned to null and validation has to be done in order to fulfill the requirement of mandatory keys.  However. The issue with the the default values are still problem as they are assigned null by terraform as long as the object is specified. 

### Proposal

My initial use case is the optimum solution. Explanation:

If type is declared as object and it contains optional attributes

- Allow `default` to only contain optional attributes.

During processing of the variable:

1. if any optional attributes are missing in the input and a default exists for those attributes, add them to the input
2. then evaluate the input against the requirement (type specification)
3. then assume the attribute is valid (with the defaults applied)

My rationale:
* Clean and intuitive code
* No unnecessary coding to meet the requirement

_If you feel that support for more complex structures hinders implementing of the solution, then my advise would be to only support a reduced number of levels, such as minimum one_

### References

_No response_

locker95:
Variable-level `default` has to satisfy every required attribute on an object type, so you cannot use it to “fill in” only the optional ones. That belongs on `optional(type, default)` in the type constraint.

Added a short note under the `default` argument with a small example and a link to the optional object attributes docs.

Fixes #727

### DIFF
--- a/content/terraform/v1.12.x/docs/language/block/variable.mdx
+++ b/content/terraform/v1.12.x/docs/language/block/variable.mdx
@@ -112,6 +112,32 @@ variable "region" {
 
 If you specify both the `type` and `default` arguments, the default value must convert to the specified type. The `default` argument requires a literal value and cannot reference other objects in the configuration.
 
+If the variable type is an object (or a collection of objects), a `default`
+on the **variable** still has to satisfy every **required** attribute in that
+type. You cannot use the variable-level `default` to supply only the optional
+fields.
+
+To give optional attributes fallback values when a caller omits them, set the
+default on the attribute itself with
+[`optional(type, default)`](/terraform/language/expressions/type-constraints#optional-object-type-attributes)
+in the type constraint:
+
+```hcl
+variable "db" {
+  type = object({
+    name                = string
+    user_name           = string
+    deletion_protection = optional(bool, true)
+    backup              = optional(bool, true)
+  })
+}
+```
+
+With that type, a caller can pass `{ name = "app", user_name = "app" }` and
+Terraform fills in `deletion_protection` and `backup`. See
+[Optional Object Type Attributes](/terraform/language/expressions/type-constraints#optional-object-type-attributes)
+for nested structures and more examples.
+
 #### Summary
 
 - Data type: Expression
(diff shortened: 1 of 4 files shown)


crw:

PR #3045, opened 2026-08-08
docs(terraform): clarify map(...) in for_each is a type constructor

### LINKED ISSUE #740: Docs: map function
### Terraform Version

```shell
$ terraform version
Terraform v1.5.2
on linux_amd64

is 1.6.3. You can update by downloading from https://www.terraform.io/downloads.html
```

### Affected Pages

https://developer.hashicorp.com/terraform/language/meta-arguments/for_each

### What is the docs issue?

The page contains this example:

```
variable "vpcs" {
  type = map(object({
    cidr_block = string
  }))
}
```

It appears to be using a function called `map`.  Which leads to:  

 https://developer.hashicorp.com/terraform/language/functions/map

"The map function is no longer available. Prior to Terraform v0.12..."

### Proposal

Should the example be rewritten without `map`? (It is also referencing `object`,  does that need to be updated also?  not sure.) Thanks.

### References

_No response_

locker95:
The for_each chaining example uses `type = map(object({...}))`. That looks a lot like the removed `map()` function page, so people click through and hit the deprecation note.

Added a short comment that this is the map type constructor.

Fixes #740

### DIFF
--- a/content/terraform/v1.12.x/docs/language/meta-arguments/for_each.mdx
+++ b/content/terraform/v1.12.x/docs/language/meta-arguments/for_each.mdx
@@ -106,6 +106,8 @@ internet gateway for each VPC:
 
 ```hcl
 variable "vpcs" {
+  # map(...) here is a type constructor (a map of objects), not the removed
+  # map() function. See type constraints for details.
   type = map(object({
     cidr_block = string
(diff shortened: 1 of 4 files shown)


crw:

PR #3047, opened 2026-08-08
docs(terraform): explain connection agent_identity selection

### LINKED ISSUE #737: Clarify Documentation for agent_identity remote-exec parameter.
### Terraform Version

```shell
Terraform v1.6.6
on darwin_arm64
```

### Affected Pages

https://developer.hashicorp.com/terraform/language/resources/provisioners/connection
Screenshot 2024-01-02 at 10 24 15 PM

### What is the docs issue?

I was troubleshooting some ssh connectivity issues, and it was unclear what the `agent_identity` parameter actually does.  I thought perhaps terraform found a way to request a specific cert by it's fingerprint from the ssh-agent. Upon reading the source code, I saw these extremely helpful comments that explain how this feature works. https://github.com/hashicorp/terraform/blob/6114ec4390db6d419b3302a3de87a39f99a233dc/internal/communicator/ssh/provisioner.go#L540

Note: I would love to see the output of the sorted list in the DEBUG console logs so that I know which identities are being tried and in what order. In high security (gov) environments, many servers fail with too many connection attempts after just one try.

### Proposal

I found this comment to be useful
`
// sortSigners moves a signer with an agent comment field matching the
// agent_identity to the head of the list when attempting authentication. This
// helps when there are more keys loaded in an agent than the host will allow
// attempts.
`

Rewriting it a bit and putting it in the argument reference table here: https://developer.hashicorp.com/terraform/language/resources/provisioners/connection 
seems like a good idea. 

Perhaps something like: 

Sets the preferred identity from the ssh agent for authentication. When Terraform gets a list of certificates from an agent, it sorts that list placing certificates matching `agent_identity` first in the list to avoid failures when there are more certificates in the list than the host will allow connection attempts.

### References

_No response_

locker95:
`agent_identity` was only described as a preferred identity from the agent. Expanded it: path to a key file (or matching agent comment), used to try that agent key first when many are loaded.

Fixes #737


crw:

PR #3048, opened 2026-08-08
docs(terraform): clarify -filter paths are from the configuration root

### LINKED ISSUE #723: Improve documentation for `terraform test`
### Terraform Version

```shell
Terraform v1.10.4
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v5.94.0

is 1.11.3. You can update by downloading from https://www.terraform.io/downloads.html
```

### Affected Pages

https://developer.hashicorp.com/terraform/cli/commands/test#terraform-test-command

### What is the docs issue?

The current documentation for `terraform test` is not explicit enough with the example usage, especially for the optional parameters such as `-filter`. The docs mention:

[-filter=testfile](https://developer.hashicorp.com/terraform/cli/commands/test#filter-testfile) - Limits the terraform test operation to the specified test files.

However the above doesn't work. With a set of tests within a recommended `/tests` directory with filenames such as `01_test.tftest.hcl`, `02_test.tftest.hcl` etc, running `terraform test -filter=02_test.tftest.hcl` or even `terraform test -filter="02_test.tftest.hcl"` does not work:

```hcl
❯ terraform test -filter=02_test.tftest.hcl

Success! 0 passed, 0 failed.
```

```hcl
❯ terraform test -filter="02_test.tftest.hcl"

Success! 0 passed, 0 failed.
```

Instead, you must write the filter as a string, and include the full path to the file, including the `/tests` directory. 

```hcl
❯ terraform test -filter="tests/02_test.tftest.hcl"
tests/02_test.tftest.hcl... in progress
  run "unit_test"... pass
  run "e2e_test"... pass
tests/02_test.tftest.hcl... tearing down
tests/02_test.tftest.hcl... pass
```

I could understand this if using a non-standard directory name for your tests, however this is not expected if using the default recommended naming of `/tests`. One would assume that if `terraform test` automatically uses test files (`*.tftest.hcl`) that are contained in a directory with this name, then the simple filter should function the same way.

### Proposal

Please update the docs with more examples and explicitly mention the way you must use filters to reference specific test files, and call out that it's required to directly reference the directory name, even if it is the default value. In the longterm, perhaps it would be possible to omit having to pass in the directory name in the filter if using the default naming of `/tests` 🙂 

### References

[terraform test documentation](https://developer.hashicorp.com/terraform/cli/commands/test#terraform-test-command)

locker95:
`-filter` matches paths from the configuration root (where you run `terraform test`), not just the basename under `tests/`. If you pass only `02_test.tftest.hcl`, Terraform finds nothing and reports 0 tests.

Docs now say to use `tests/name.tftest.hcl` for files in the default directory, and call that out next to the examples.

v1.12.x–v1.16.x (beta).

Fixes #723

### DIFF
--- a/content/terraform/v1.12.x/docs/cli/commands/test.mdx
+++ b/content/terraform/v1.12.x/docs/cli/commands/test.mdx
@@ -32,7 +32,7 @@ The following options apply to the Terraform `terraform test` command:
 
 * `-cloud-run=<module source>` - This test run executes remotely on HCP Terraform within the specified Terraform [private registry](/terraform/language/modules/sources#terraform-registry) module.
 
-* `-filter=testfile` - Limits the `terraform test` operation to the specified test files.
+* `-filter=testfile` - Limits the `terraform test` operation to the specified test file(s). Paths are relative to the **configuration root** (the directory where you run `terraform test`), not relative to the testing directory alone. For a file under the default `tests/` directory, pass `tests/name.tftest.hcl` (for example `-filter=tests/02_test.tftest.hcl`). Passing only the basename (such as `-filter=02_test.tftest.hcl`) does not match files inside `tests/` and Terraform runs zero tests.
 
 * `-json` - Displays machine-readable JSON output for your testing results.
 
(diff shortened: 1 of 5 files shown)


crw:

PR #3052, opened 2026-08-08
docs(vault): document kv metadata patch

### LINKED ISSUE #30954: [Doc] Vault kv metadata doc page doesn't contain patch subcommand
Hi,

The current Vault binary supports `vault kv metadata patch` which we use to partially update the custom metadata.

However I found that the patch subcommand is not mentioned in  [CLI / kv / metadata](https://developer.hashicorp.com/vault/docs/commands/kv/metadata).

Is it missed, or described in somewhere else in the docuementation?

locker95:
The CLI supports `vault kv metadata patch` (partial updates / remove individual custom_metadata keys), but the commands page only listed delete/get/put. Added the subcommand to the list and a short section with examples for v1.19–v1.21 and v2.x.

Fixes hashicorp/vault#30954


crw:

PR #3055, opened 2026-08-08
docs(vault): PKI role ttl accepts full duration strings including days

### LINKED ISSUE #31404: docs: create/update role allows days as a time suffix

locker95:
Create/Update Role still said hour was the largest suffix for `ttl` / `max_ttl`. That is wrong — Vault's duration format supports `d` (days) and the other units on the duration-format page (and people commonly set 90d cert lifetimes).

Replaced the old wording with a pointer to that page on recent PKI API docs.

Fixes hashicorp/vault#31404

### DIFF
--- a/content/vault/v1.19.x/content/api-docs/secret/pki/index.mdx
+++ b/content/vault/v1.19.x/content/api-docs/secret/pki/index.mdx
@@ -3189,13 +3189,13 @@ request is denied.
 
 - `ttl` `(string: "")` - Specifies the Time To Live value to be used for the
   validity period of the requested certificate, provided as a string duration
-  with time suffix. Hour is the largest suffix. The value specified is strictly
+  using [duration format strings](/vault/docs/concepts/duration-format) (hours, days, etc.). The value specified is strictly
   used for future validity. If not set, uses the system default value or the
   value of `max_ttl`, whichever is shorter. See `not_after` as an alternative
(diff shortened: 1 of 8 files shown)


crw:

PR #3056, opened 2026-08-09
docs(vault): clarify sign-intermediate TTL vs issuer expiry

### LINKED ISSUE #31403: docs: `sign-intermediate`'s behavior with notAfter beyond the signer's expiry doesn't match the docs
**Describe the bug**

According to `sign-intermediate`'s [`ttl`](https://developer.hashicorp.com/vault/api-docs/secret/pki#ttl-2), "this can be after the expiration of the signing CA". However, even when [`leaf_not_after_behavior`](https://developer.hashicorp.com/vault/api-docs/secret/pki#leaf_not_after_behavior) is set to `permit`, a `notAfter` beyond the expiry of the signer will get truncated. While I suppose technically a too-long TTL *can* indeed be given, I don't think anybody reading that description would expect truncation.

Also, according to the [`sign-intermediate`](https://developer.hashicorp.com/vault/api-docs/secret/pki#sign-intermediate) docs, that endpoint does not accept the `enforce_leaf_not_after_behavior` option (per docs, only `sign-verbatim` does). However, in practice and based on [code examination](https://github.com/hashicorp/vault/blob/main/builtin/logical/pki/path_root.go#L396C38-L396C69), that option does seem to be accepted (and setting it to true and the [`leaf_not_after_behavior`](https://developer.hashicorp.com/vault/api-docs/v1.16.x/secret/pki#leaf_not_after_behavior) to `permit` seems to be the only way to get what appears to have been the 1.16 default behavior of allowing certs extending past the CA's validity -- neither option on their own seems to do it, which I find weird).

It's possible this is a code bug rather than a doc bug -- certainly I find `enforce_leaf_not_after_behavior` to be a bit odd, and also really wish I could make a single request with a "yes, I mean it, just sign this" flag -- but it looks like a deliberate choice where people just forgot to update the docs.

locker95:
`sign-intermediate`'s `ttl` docs said the cert can live past the signing CA without much nuance. By default Vault still truncates `NotAfter` to the issuer; you only get a longer intermediate if you wire up `enforce_leaf_not_after_behavior` and the issuer's `leaf_not_after_behavior` (`permit`, etc.).

Updated the parameter text to match that.

Fixes hashicorp/vault#31403

### DIFF
--- a/content/vault/v1.19.x/content/api-docs/secret/pki/index.mdx
+++ b/content/vault/v1.19.x/content/api-docs/secret/pki/index.mdx
@@ -734,9 +734,18 @@ intermediary goes beyond the prescribed length.
 
 - `ttl` `(string: "")` - Specifies the requested Time To Live (after which the
   certificate will be expired). This cannot be larger than the engine's max (or,
-  if not set, the system max). However, this can be after the expiration of the
-  signing CA. See `not_after` as an alternative for setting an absolute end date
-  (rather than a relative one).
+  if not set, the system max). See `not_after` as an alternative for setting an
+  absolute end date (rather than a relative one).
+
+  By default, if the requested `NotAfter` would extend past the signing
+  issuer's expiry, Vault **truncates** it to the issuer's `NotAfter` rather
+  than allowing a longer intermediate. To control that behavior, set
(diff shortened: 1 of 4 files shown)


crw:

PR #3068, opened 2026-08-10
docs(nomad): clarify job ID uniqueness vs optional name

### LINKED ISSUE #20149: Docs: Nomad job name vs job ID confusion
### Nomad version
1.7.5

### Operating system and Environment details
Debian 12

### Issue
Docs at https://developer.hashicorp.com/nomad/docs/job-specification/job say that `Job names are unique per region or namespace.`
which does not seem to be accurate as one can have multiple jobs with identical `name=` attribute in the same region and namespace.

locker95:
The intro on the job block page said job *names* are unique per region/namespace. That mixes up the job ID (`job "…"`) with the optional `name` field.

Job IDs are what must be unique. `name` is just a display string and can be the same on different jobs.

Fixes hashicorp/nomad#20149

### DIFF
--- a/content/nomad/v1.10.x/content/docs/job-specification/job.mdx
+++ b/content/nomad/v1.10.x/content/docs/job-specification/job.mdx
@@ -17,7 +17,7 @@ last_modified: 2025-10-10T11:41:43-05:00
 The `job` block is the top-most configuration option in the job specification.
 A job is a declarative specification of tasks that Nomad should run. Jobs have
 one or more task groups, which are themselves collections of one or more tasks.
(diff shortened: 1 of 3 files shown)


tgross (changes requested):
Just change the "Job names are unique" to Job IDs are unique". All the rest of this is fluff and the Job name field is already clear (and users are guided away from using it by the UX). You've also not backported to the supported 1.10.x version

locker95:
yeah, overdid it. cut it down to just changing names→IDs on 1.10/1.11/2.0.

crw:

PR #3069, opened 2026-08-10
docs(nomad): explain 4647 RPC vs 4648 Serf for cluster join

### LINKED ISSUE #20119: Not clear documentation: why servers and clients uses different port to join server 4647 VS 4648
image

1. This not clear information is documented on this page: https://developer.hashicorp.com/nomad/tutorials/manage-clusters/clustering#manual-clustering

2. Also please clarify 'The port corresponds to the RPC port. ' part. Which RPC port do you mean here: nomad, consul etc? Provide a link to corresponding configuration option.

3. > This means only one server must be specified because, after initial contact, the full set of servers in the client's region are shared with the client.

   I do not remember where but I am sure that you used array of servers as the value for this option. Anyway we can see that you require array. Notice `[]` for the value. It is ambiguous is `only one server must be specified` why this one value is passed via array?

locker95:
The manual clustering page showed servers on `:4648` and clients on `:4647` without saying why, and the “RPC port” note only covered the client path. Also clarified that `retry_join` is a list (multiple seeds ok) even though one working server is enough for clients.

Fixes hashicorp/nomad#20119

### DIFF
--- a/content/nomad/v1.11.x/content/docs/deploy/clusters/connect-nodes.mdx
+++ b/content/nomad/v1.11.x/content/docs/deploy/clusters/connect-nodes.mdx
@@ -47,7 +47,7 @@ server {
   enabled          = true
   bootstrap_expect = 3
 
-  # This is the IP address of the first server provisioned
+  # First server address; :4648 is the Serf (gossip) port for server membership
   server_join {
     retry_join = ["<known-address>:4648"]
   }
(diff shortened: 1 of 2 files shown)


content/nomad/v1.11.x/content/docs/deploy/clusters/connect-nodes.mdx
```diff
@@ -47,7 +47,7 @@ server {
   enabled          = true
   bootstrap_expect = 3
 
-  # This is the IP address of the first server provisioned
+  # First server address; :4648 is the Serf (gossip) port for server membership
```
tgross, line 50: `:4648` is not a port number, and not editing your LLM output to realize that putting a `;` right before the `:` is particularly egregious 😁 This can just be "# This is the IP address and gossip port of the first server provisioned`

content/nomad/v1.11.x/content/docs/deploy/clusters/connect-nodes.mdx
```diff
@@ -81,13 +81,23 @@ The client node's server list can be updated at run time using the
...
-The port corresponds to the RPC port. If no port is specified with the IP
-address, the default RPC port of `4647` is assumed.
+Servers and clients use different default ports when joining:
 
-As servers are added or removed from the cluster, this information is pushed to
-the client. This means only one server must be specified because, after initial
-contact, the full set of servers in the client's region are shared with the
-client.
+- **Server-to-server** `retry_join` / `nomad server join` uses Nomad's
+  [Serf](/nomad/docs/configuration#ports) (gossip) port, default **`4648`**
```
tgross, line 87: Stop bolding everything.

You've removed most of the original test in lieu of a bunch of repetitive description of this port. Undo that.

tgross (commented):
You didn't update the config join docs as mentioned in the issue. This PR in general lacks a human editor's touch, so I'm inclined to reject it outright.

locker95:
ack — I'll fix the port comment, put the original wording back, and actually update the config join docs the issue asked for.

crw:

PR #3070, opened 2026-08-10
docs(nomad): clarify job dispatch takes a registered job ID

### LINKED ISSUE #16052: docs: add clarifying example to job dispatch command docs
https://developer.hashicorp.com/nomad/docs/commands/job/dispatch

The docs here don't make it clear that `nomad job dispatch <name>` is referring to JobID (as opposed to the job filename). It's also not clear that the job must be "run" first before "dispatch". Let's fixup the docs so these points are clear.

locker95:
The dispatch page already used job IDs in the examples, but the usage blurb still read like you might pass a job file. Spelled out that the argument is the registered parameterized job ID and you need `nomad job run` (or the API) first.

Fixes hashicorp/nomad#16052

### DIFF
--- a/content/nomad/v1.11.x/content/commands/job/dispatch.mdx
+++ b/content/nomad/v1.11.x/content/commands/job/dispatch.mdx
@@ -27,11 +27,18 @@ programming languages.
 nomad job dispatch [options] <parameterized job> [input source]
 ```
 
-`<parameterized job>` is the **job ID** of a parameterized job already
-registered in Nomad (the string in `job "…"`), not a path to a job file.
-Register the parent job first with [`nomad job run`](/nomad/commands/job/run)
-(or the equivalent API). Dispatch only creates instances of that registered
-job; it does not submit a new job specification from disk.
+`<parameterized job>` is the job ID of a parameterized job already
+registered with Nomad (not a path to a job file). Register it first with
+`nomad job run`, then dispatch:
+
+```shell-session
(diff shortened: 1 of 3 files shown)


content/nomad/v1.11.x/content/commands/job/dispatch.mdx
```diff
@@ -27,10 +27,15 @@ programming languages.
 nomad job dispatch [options] <parameterized job> [input source]
 ```
 
-Dispatch creates an instance of a parameterized job. A data payload to the
-dispatched instance can be provided via stdin by using "-" for the input source
-or by specifying a path to a file. Metadata can be supplied by using the meta
-flag one or more times.
+`<parameterized job>` is the **job ID** of a parameterized job already
```
tgross: Stop bolding everything. Also, the user doesn't need to have an example spelling out where to find the job ID in the jobspec that they already submitted. Let's have some respect for the intelligence of the reader.

tgross (changes requested):
This doesn't actually solve the issue described in https://github.com/hashicorp/nomad/issues/16052 which is that the example is incomplete. Most of the rest of this is noise.

You've also not completed backports to supported versions.

locker95:
yep, I'll trim the noise and make the dispatch example complete on the supported versions instead.

locker95:
rewrote this around a short run→dispatch example and backported to 1.10 as well. dropped the bold lecture.

crw:

PR #3074, opened 2026-08-10
feat: Adds Boundary to content exclusion transform

Dan-Heath:
Adds Boundary to the content exclusion transform map:

* Maps a directive tag (the text before the colon in a BEGIN/END comment) to
 * the product slug the directive is scoped to. To support a new product, add a
 * tag here - no new processor file is required.


williamdalessandro:
Hey @Dan-Heath if you're trying to add content exclusion support to Boundary make sure to follow the steps [here](https://github.com/hashicorp/web-unified-docs/tree/main/scripts/prebuild/mdx-transforms/exclude-content#adding-a-new-product) - just looks like you're missing a single step. Also copy and pasting an existing test case and replacing it with "Boundary" to see content exclusion in action would be helpful, but not totally necessary

Dan-Heath:
> Hey @Dan-Heath if you're trying to add content exclusion support to Boundary make sure to follow the steps [here](https://github.com/hashicorp/web-unified-docs/tree/main/scripts/prebuild/mdx-transforms/exclude-content#adding-a-new-product) - just looks like you're missing a single step. Also copy and pasting an existing test case and replacing it with "Boundary" to see content exclusion in action would be helpful, but not totally necessary

Thank you @williamdalessandro ! I think I got the missing step and a test case added.

williamdalessandro (approved):
LGTM!

PR #3101, opened 2026-08-12
202609 agentic merge main

VioletHynes:
Merging main to `vault/202609-agentic` as per Slack thread


schavis:
The EDU folks keep the assembly/release branches synced with main. You shouldn't need to create a PR for that purpose

PR #3109, opened 2026-08-13
Hvd misc bug fixes

williamdalessandro:
Fixing some leftover issues with the hvd migration- fixing search issue that came as a result of the unique setup of both HVD and WAF. 

Since the folder organization of these products have 2 nested `docs/` folders and since the algolia search reads the from the file location on disk (i.e. `public/content/validated-designs/docs/docs/vault/installation-guide/vms/introduction.mdx`) rather than through the nextjs routing (i.e. `pages/validated-designs/[...page].tsx`, which pulls out the page params as `vault/installation-guide/vms/introduction`) , a special case has to be created here. 

The authoring convention used here for HVD and WAF puts .mdx files under `docs/docs/` so that non-MDX operational files (AGENTS.md, templates, etc.) can live at `docs/` without being served as content.

Updated tests accordingly


content/validated-designs/docs/docs/terraform/installation-guide/architecture.mdx
```diff
@@ -1,9 +1,9 @@
 ---
-page_title: Architecture
+page_title: Architecture | Terraform Installation Guide
```
danbr-ibm:
```suggestion
page_title: Architecture
```

Let's back this change out for now and we can address all the page_titles in one go.

danbr-ibm (dismissed):

ryan-yu-ibm (dismissed):
Changes make sense to me ✅

danbr-ibm (approved):

ryan-yu-ibm (approved):

PR #3121, opened 2026-08-16
Fix broken component configuration link in Stacks tfcomponent docs

### LINKED ISSUE #1247: broken link: content/terraform/v1.13.x/docs/language/block/stack/tfcomponent/index.mdx
Link on [this line](https://github.com/hashicorp/web-unified-docs/blob/b224edd/content/terraform/v1.13.x/docs/language/block/stack/tfcomponent/index.mdx?plain=1#L16) is not a valid link. I couldn't determine with confidence what the correct link should be.

Gaalbu:
Fixes #1247

## What was wrong

In `content/terraform/v1.13.x/docs/language/block/stack/tfcomponent/index.mdx` (line 16 as referenced in the issue), the link text "defining Stack component configuration files" pointed to:

```
(terraform/language/stacks/component/config)
```

This is a **relative** path (missing the leading `/`), while every other link in this repo's `.mdx` content uses absolute paths rooted at the product (e.g. the very same file's other links: `/terraform/language/block/stack/tfcomponent/component`, `/terraform/language/block/locals`, etc.). Without the leading slash, the link resolves relative to the current page instead of to the intended top-level route, producing a broken/invalid link.

## How I determined the correct target

The issue reporter noted they couldn't confidently identify the intended destination, so I verified it two ways:

1. **The target page exists in this repo.** `content/terraform/v1.13.x/docs/language/stacks/component/config.mdx` exists, with `page_title: Define configuration` and content that opens with "In the component configuration file, you declare what infrastructure components are part of the Stack" — this is precisely what the link text ("defining Stack component configuration files") describes.
2. **The live docs confirm the route.** `https://developer.hashicorp.com/terraform/language/stacks/component/config` renders that same "Define configuration" page, confirming `/terraform/language/stacks/component/config` is the correct, currently-live absolute path.

So the only defect was the missing leading `/`; the path segments themselves were already correct.

## Fix

Changed:
```diff
-Learn more about [defining Stack component configuration files](terraform/language/stacks/component/config).
+Learn more about [defining Stack component configuration files](/terraform/language/stacks/component/config).
```

The identical broken link (copy-pasted per-version, as is this repo's content structure) also existed in the `v1.14.x`, `v1.15.x`, and `v1.16.x (beta)` copies of this same file. I fixed all four in this PR to avoid leaving three more instances of the same known bug for another link-checker report — this repo's own history shows this cross-version-folder pattern for single-bug fixes (e.g. the vault `migration-checklist.mdx` fixes applied across `v1.18.x`–`v2.x` in one PR each).

## Testing

Manually verified the corrected route resolves on the live site (`developer.hashicorp.com/terraform/language/stacks/component/config`) and matches the link text's intent. The repo's `broken-link` check (`scripts/broken-link.mjs`) requires a local Docker-based site build, which was out of scope for this one-line, four-file link fix; the fix follows the exact absolute-path convention used by every other link in the same file.


trujillo-adam (dismissed):
Thanks for fixing this!

Gaalbu:
Rebased on latest main to resolve the merge conflict (v1.16.x moved out of RC). Link fix applied to v1.13.x–v1.16.x.

PR #3145, opened 2026-08-19
Update metrics documentation for Vault cluster behavior

AkashGuptaHC:
Clarified behavior of metric requests in Vault clusters regarding active and standby nodes.

### DIFF
--- a/content/vault/v2.x/content/api-docs/system/metrics.mdx
+++ b/content/vault/v2.x/content/api-docs/system/metrics.mdx
@@ -37,10 +37,10 @@ $ curl \
   --header "X-Vault-Token: ..." \
     'http://127.0.0.1:8200/v1/sys/metrics?format=prometheus'
 ```
-When requesting metric information, only the active node in a Vault cluster
-responds with data. Follower nodes ignore metric requests, which typically
-results in a timeout response. For example, `curl` command returns with the
-error code `0` and no output.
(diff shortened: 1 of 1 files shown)


AkashGuptaHC:
Reference: https://ibm-hashicorp.slack.com/archives/C09L31FC7CL/p1787025529737919

content/vault/v2.x/content/api-docs/system/metrics.mdx
```diff
@@ -37,10 +37,10 @@ $ curl \
   --header "X-Vault-Token: ..." \
     'http://127.0.0.1:8200/v1/sys/metrics?format=prometheus'
 ```
-When requesting metric information, only the active node in a Vault cluster
-responds with data. Follower nodes ignore metric requests, which typically
-results in a timeout response. For example, `curl` command returns with the
-error code `0` and no output.
+When requesting metric information, only the active & the performance standby nodes in a Vault cluster
+respond with data. Just a standby node will ignore the metric requests, which typically
```
ncabatoff, line 41: That isn't what I observe.  When I try it, I get a response from the standby node, which I assume is based on forwarding, whether I use curl or the vault CLI.

schavis:
Per @ncabatoff 's comments, it looks like this PR is technically incorrect so I'm going to close it.

PR #3148, opened 2026-08-19
docs: clarify generate-config-out limitation with for_each resources

soepic1:
Context: hashicorp/terraform#38032

While importing an existing Aurora cluster into a for_each-managed Terraform module, I found that -generate-config-out produces isolated, single-resource configuration blocks rather than a consolidated for_each block, with no documented workaround. This PR adds a note explaining the temporary-address workaround, referencing the related discussion in #38032 for further context.

### DIFF
--- a/content/terraform/v1.5.x/docs/language/import/generating-configuration.mdx
+++ b/content/terraform/v1.5.x/docs/language/import/generating-configuration.mdx
@@ -148,3 +148,7 @@ $ terraform plan -generate-config-out=generated.tf
 ```
 
 In the example above, Terraform still generates configuration and writes it to `generated.tf`. This error stems from a conflict between the `ipv6_address_count` and `ipv6_addresses` arguments. The resource supports both of these arguments, but you must choose only one when configuring the resource. You could fix the error by removing one of these two arguments, then running `terraform plan` again to check that there are no further issues.
+
+### Importing resources with `for_each`
+
+~> **Note:** When importing a resource that will ultimately be managed with `for_each`, `-generate-config-out` cannot generate configuration directly against a `for_each`-keyed address (for example, `aws_s3_bucket.this["my-bucket"]`). Instead, generate the configuration against a temporary, non-indexed resource address first, then manually refactor the generated block into your `for_each` structure afterward. See [hashicorp/terraform#38032](https://github.com/hashicorp/terraform/issues/38032) for further context and discussion.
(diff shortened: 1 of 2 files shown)


content/terraform/v1.5.x/docs/language/import/generating-configuration.mdx
```diff
@@ -148,3 +148,7 @@ $ terraform plan -generate-config-out=generated.tf
 ```
 
 In the example above, Terraform still generates configuration and writes it to `generated.tf`. This error stems from a conflict between the `ipv6_address_count` and `ipv6_addresses` arguments. The resource supports both of these arguments, but you must choose only one when configuring the resource. You could fix the error by removing one of these two arguments, then running `terraform plan` again to check that there are no further issues.
+
+### Importing resources with `for_each`
+
+~> **Note:** When importing a resource that will ultimately be managed with `for_each`, `-generate-config-out` cannot generate configuration directly against a `for_each`-keyed address (for example, `aws_s3_bucket.this["my-bucket"]`). Instead, generate the configuration against a temporary, non-indexed resource address first, then manually refactor the generated block into your `for_each` structure afterward. See [hashicorp/terraform#38032](https://github.com/hashicorp/terraform/issues/38032) for further context and discussion.
```
trujillo-adam, line 154:
```suggestion
The `-generate-config-out` flag cannot generate configuration directly against a `for_each`-keyed address, for example `aws_s3_bucket.this["my-bucket"]`. 

When importing a resource that you intend to manage with code containing a `for_each` argument, generate the configuration against a temporary, non-indexed resource address first, then manually refactor the generated block into your `for_each` structure afterward. 

Refer to [hashicorp/terraform#38032](https://github.com/hashicorp/terraform/issues/38032) for more information.
```

trujillo-adam (commented):
Thanks for adding this! I added a few edits for compliance with our style guide. 
Also, looking at the linked issue, it seems that this behavior has existed since 1.14. I am unsure if it is fixed in 1.16. In any case, please replicate this change to the  1.14 version of the page. 

Let us know when you've addressed the feedback and we can approve.

content/terraform/v1.5.x/docs/language/import/generating-configuration.mdx
soepic1: Thanks @trujillo-adam! I've updated the wording to match your style edits across both v1.5.x and v1.14.x. Ready for another review!"

PR #3176, opened 2026-08-21
docs: Style guide updates

Dan-Heath:
Updates the style guide with 2 new best practices:

- Do not use "this" as a pronoun
- Use inline links sparingly

It also fixes some pre-existing typos.

### DIFF
--- a/docs/style-guide/general/language.md
+++ b/docs/style-guide/general/language.md
@@ -47,14 +47,14 @@ Refer to general roles, such as a "developer", "engineer", or "administrator", i
 - **keywords**: writing, word choice, pronouns, ambiguity
 - **content sets**: docs, tutorials, WAF, certifications
 
-Using “this” in place of a noun can create ambiguous sentences that confuse users about who or what is performing the action.
+Using “this” in place of a noun can create ambiguous sentences that confuse users about who or what is performing the action. If you use "this", make sure it has a clear noun referent.
 
 ### Examples
 
 **Do:**
 
 - `Update the worker configuration file and restart the worker. Restarting the worker ensures that the new authentication settings take effect.`
-- `Create a user account and assign the user to an administrator role. The administrator role grants access to manage targets.`
+- `Create a user account and assign the user to an administrator role. This role grants access to manage targets.`
 
 **Don't:**
 
--- a/docs/style-guide/general/language.md
+++ b/docs/style-guide/general/language.md
@@ -47,14 +47,14 @@ Refer to general roles, such as a "developer", "engineer", or "administrator", i
 - **keywords**: writing, word choice, pronouns, ambiguity
 - **content sets**: docs, tutorials, WAF, certifications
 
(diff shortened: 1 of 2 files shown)


docs/style-guide/general/language.md
```diff
@@ -42,9 +42,28 @@ Refer to general roles, such as a "developer", "engineer", or "administrator", i
 
 - `Provide a token to the developer so that he can deploy services using the API.`
 
+## Do not use "this" as a pronoun
+
+- **keywords**: writing, word choice, pronouns, ambiguity
+- **content sets**: docs, tutorials, WAF, certifications
+
+Using “this” in place of a noun can create ambiguous sentences that confuse users about who or what is performing the action. If you use "this", make sure it has a clear noun referent.
```
schavis:
```suggestion
Using “this” in place of a noun can create ambiguous sentences that confuse users about who or what is performing the action.
```
I don't think we should offer exceptions to the rule, especially as we have more folks using AI, for which explicit rules typically generate better results. I have yet to run into a situation where I couldn't suggest a rewrite to avoid "this" without harming the sentence clarity.

docs/style-guide/general/language.md
```diff
@@ -42,9 +42,28 @@ Refer to general roles, such as a "developer", "engineer", or "administrator", i
...
+
+### Examples
+
+**Do:**
+
+- `Update the worker configuration file and restart the worker. Restarting the worker ensures that the new authentication settings take effect.`
+- `Create a user account and assign the user to an administrator role. This role grants access to manage targets.`
```
schavis:
```suggestion
- `Create a user account and assign the user to an administrator role. The admin role grants access to manage targets.`
```
I'm guessing you didn't mean for the "Do" example to be identical to the "Don't" example?

docs/style-guide/general/language.md
```diff
@@ -178,7 +197,7 @@ The following list contains common words and phrases that you should avoid in ed
 | carte blanche | full permission, admin access | French word that literally translates to "blank document" and means "unlimited authority". |
 | via | Choose a more concrete word to describe the relationship. | Latin word meaning "by way of". |
 | vice versa | conversely | Latin word meaning "the other way around". |
-| sanity check | preliminary check, verification, dry run | jarbon, ableist |
+| sanity check | preliminary check, verification, dry run | jargon, ableist |
```
schavis, line 200: Good call! Jargon definitely shouldn't be allowed, but we use jarbon all the time, so we should definitely allow it 😆

docs/style-guide/general/language.md
Dan-Heath: Good point. I debated add that sentence, I'm fine with removing it.

docs/style-guide/general/language.md
Dan-Heath: Thank you!

aimeeu (approved):

PR #3216, opened 2026-08-25
add entra id to agentic IAM guide

mister-ken:
Add Entra ID to agentic IAM.

- Added [Entra ID agentic IAM](https://unified-docs-frontend-preview-9gxdedo06-hashicorp.vercel.app/vault/ai/iam/providers/entraid) and supporting images
- Updated nav data json
- Removed beta tags for verify
- Fixed some typos on verify


content/vault/v2.x/content/ai/iam/providers/entraid.mdx
```diff
@@ -0,0 +1,350 @@
...
+      SECRET_VALUE=<YOUR SECRET VALUE>
+   ```
+
+## Step 6: Configure Vault for Entra ID
+
+Use the values gathered from Microsoft Entra ID to configure an OAuth resource server profile in Vault that validates agent JWTs.
+
+<Note>
+
+   Microsoft Entra ID JWTs use `uti` as the unique identifier claim instead of the standard `jti`. Set `unique_id_claim` to `uti` and set `optional_authorization_details` to `false` to make the `authorization_details` claim optional in issued JWTs.
```
davidadeleon:
```suggestion
   Microsoft Entra ID JWTs use `uti` as the unique identifier claim instead of the standard `jti`. Set `unique_id_claim` to `uti` and set `optional_authorization_details` to `true` to make the `authorization_details` claim optional in issued JWTs.
```

davidadeleon (commented):
The only thing I would probably want to see is just consistency when talking about entity IDs and entity-alias IDs. I know everything is generally prefaced with it being example output, but having the IDs be consistent across the examples helps me reason around what is what. For instance we don't show the complete entity ID being returned from the creation statement. The shortened snippet that is created doesn't match the canonical ID provided in the entity-alias creation call, or the read call. It just makes tracking the relationship between these objects tougher. One thing I've seen done in other examples throughout our documentation is to pull the needed values from the calls into ENV vars that can be re-used throughout the other examples.

content/vault/v2.x/content/ai/iam/providers/entraid.mdx
```diff
@@ -0,0 +1,362 @@
+---
+layout: docs
+page_title: Configure Microsoft Entra ID for agent IAM
+description: >-
+   Use Microsoft Entra ID and Vault to secure agent permissions with App registrations, OAuth resource server profiles, and entity aliases.
+---
+
+# Configure Microsoft Entra ID for agent IAM
+
+@include 'alerts/beta.mdx'
```
miagilepner: do we need the beta tag? the feature will be GA'd, but I don't know if we're considering Entra support to be beta.

mister-ken: Thanks for the heads up!

yhyakuna (approved):

PR #2938, opened 2026-07-26
Stop post condition documentation accidentally referring to `precondition` instead of `postcondition`

rewento:
## PR Summary

Fixed incorrect reference to `precondition` meta-argument in documentation of `postcondition` that introduces the required arguments of `postcondition`.

## Related Issue
<https://github.com/hashicorp/web-unified-docs/issues/2937>

PR #3223, opened 2026-08-26
fix: Stop documentation about the `postcondition` block referring to it as a `precondition` block.

SarahFrench:
Follow up to https://github.com/hashicorp/web-unified-docs/pull/2938 which didn't include the v1.16 docs.


robin-norwood (dismissed):
lgtm, looks like a merge conflict to be resolved here though.

SarahFrench:
> lgtm, looks like a merge conflict to be resolved here though.

i just fixed it - the v1.16 release caused the file I'm editing to be renamed

trujillo-adam (approved):
Thanks for fixing

robin-norwood (approved):

PR #3230, opened 2026-08-26
fix: Update command in docker compose

### LINKED ISSUE #3119: Docker Compose file for Multi-node Consul (v2.0.x) cluster is missing a flag
### Product

Consul

### Issue type

Bug

Docker compose file "consul-cluster.yml" for Multi-node Consul (v2.0.x) cluster is missing flag -client. The docker compose is suppose to start the container which accepts connections from host but without this flag the it is constrained to accept connections from 127.0.0.1 only

akshaypgore:
1. Container-1 needs flag of -client=0.0.0.0 so that the UI of consul can be accessed via local
2. Doc states to access the UI from the host and without this flag it won't be accessible

Fixes: https://github.com/hashicorp/web-unified-docs/issues/3119

### DIFF
--- a/content/consul/v2.0.x/content/docs/deploy/server/docker.mdx
+++ b/content/consul/v2.0.x/content/docs/deploy/server/docker.mdx
@@ -15,7 +15,7 @@ This topic provides an overview for deploying a Consul server when running Consu
 
 ## Deploy and run a Consul server
 
-You can start a Consul server container and configure the Consul agent from the command line. 
+You can start a Consul server container and configure the Consul agent from the command line.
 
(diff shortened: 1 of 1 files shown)


akshaypgore (commented):
The consul doc states to access UI from the host which runs containers created via docker compose. But for UI to be accessible the container-1 has to map post to host via flag -client=0.0.0.0 which was missing in command. After adding this flag the UI becomes accessible from the host which runs containers.

akshaypgore:
@boruszak Can you please help me? This PR is not being assigned to any reviewers.

danielehc (approved):

PR #3329, opened 2026-09-07
proposal: preserve created_at on content edits

nikolasrieble:
## Purpose

This PR surfaces an unexpected date-metadata behavior and proposes a small solution for team discussion. The proposal is intentionally narrow; it is not meant to assume that the policy or implementation is already settled, and we may prefer a different solution.

The behavior surfaced while reviewing [web-unified-docs-internal#2095](https://github.com/hashicorp/web-unified-docs-internal/pull/2095). The contributor did nothing wrong—the pre-commit hook behaved as implemented and staged the generated changes automatically.

## Problem observed

PR #2095 edits 32 MDX files but creates no pages. Even so, `created_at` changed in 31 files:

- **26 intentional defaults changed to repository migration dates:**

  ```diff
  -created_at: 2023-01-01T00:00:00.000Z
  +created_at: 2025-05-28T07:56:28-04:00
  ```

  The new value comes from the commit that [renamed `ptfe-releases` to `terraform-enterprise`](https://github.com/hashicorp/web-unified-docs/commit/2b02a523c361fb73083b9d4034522190d43e856d), not from the document's creation.

- **5 timestamps changed only in formatting:**

  ```diff
  -created_at: 2025-10-15T15:59:29+00:00
  +created_at: 2025-10-15T15:59:29Z
  ```

  Git 2.45 changed strict ISO UTC output from `+00:00` to `Z`. See the [Git 2.45 release notes](https://github.com/git/git/blob/master/Documentation/RelNotes/2.45.0.adoc).

## Why it happens

[#1601](https://github.com/hashicorp/web-unified-docs/pull/1601) introduced the helper for initial backfills, where generating both fields was appropriate. [#1841](https://github.com/hashicorp/web-unified-docs/pull/1841) later called the same helper from the pre-commit hook to update `last_modified`, so it also recalculates `created_at`.

## Proposed solution

When the existing hook supplies a modification date, keep the raw `created_at` from the existing metadata block before updating `last_modified`. New files and explicit history-based runs retain their existing behavior.

This is only one possible solution. In particular, feedback is welcome on whether:

- `created_at` should be immutable after initialization,
- Git history or committed frontmatter should be authoritative,
- intentional default dates should remain unchanged, and
- the pre-commit hook is the right place to maintain this metadata.

The implementation is deliberately small: 11 production lines, one adjusted test, two short documentation notes, and the three regenerated binaries required for changes under `scripts/utils/`. It does not change the hook, add CI policy, alter backfill history, or bulk-regenerate content.

## Testing

- `volta run --node 24 npx vitest run` — 295 tests passed
- `volta run --node 24 npm run coverage` — passed
- `volta run --node 24 npm run lint` — passed
- `volta run --node 24 npm run compile-prebuild` — regenerated all required binaries
- Smoke-tested against the affected TFE file: `created_at: 2023-01-01T00:00:00.000Z` remained unchanged while `last_modified` updated


LeahMarieBush (approved):
LGTM! Tested locally with a few different cases and they all worked great for me. Thanks for the improvement!

PR #3343, opened 2026-09-09
Terraform Docs: fix curl .netrc documentation links

papidb:
## What

Updates Terraform documentation links to curl's .netrc documentation from:

https://everything.curl.dev/usingcurl/netrc

to:

https://everything.curl.dev/usingcurl/netrc.html

## Why

The previous curl documentation URL no longer resolves to the intended page. The updated URL points directly to the current .netrc documentation page.

## Details

This updates the affected Terraform versioned documentation pages, including provider and module source references, so the same broken external link is fixed consistently across published Terraform docs versions.

## Validation

- Confirmed the old URL no longer appears in tracked MDX content.
- Confirmed the replacement URL resolves successfully.
- Ran Prettier checks on the touched MDX files.


radeksimko (approved):
Thanks for the fix and for applying it to all the versions.

PR #3455, opened 2026-09-17
ForwardPort: Duplicate instruction into Manual section

aimeeu:
Duplicate the `targetProduct` value instruction into the Manual section so that half-awake, overworked tech writers don't accidentally capitalize the first letter of the product name.


williamdalessandro (approved):
It looks good to me! In the preview the note is printed large so that info is definitely clearer now- thank you @aimeeu !

PR #3458, opened 2026-09-17
ForwardPort: Update forward-port-config.yml for Consul

aimeeu:
- Update Consul entry to the upcoming major Consul release.
- Fix Nomad entry
- Update instructions

Labels:
1. `forward-port:consul`
2. `forward-port:nomad`


.github/forward-port-config.yml
```diff
@@ -22,11 +22,11 @@ boundary-1.0-forward-port:
   targetVersionFolder: v1.0.x
 
 # ── Consul ───────────────────────────────────────────────────────────────────
-consul-placeholder-label:
-  sourceVersionFolder: v1.0.0
+forward-port:consul:
```
williamdalessandro: Hi @aimeeu the names here shouldn't contain a colon. These are values to reference later like from a JSON object. So the proper way to do this would be for the key name here to be anything like `consul-2.1-forward-port` and for the PR you're automatically forward porting from to have a **_label_** that is `forward-port:consul-2.1-forward-port`.

For this I'm just using a similar naming convention that Boundary does on line 18 of this file

```suggestion
consul-2.1-forward-port:
```

.github/forward-port-config.yml
williamdalessandro: I noticed the same issue here with the colon- the forward port will error out here with improper yml keys. On the PRs that are being forward ported- they should contain the label `forward-port:nomad-2.1-forward-port`

```suggestion
nomad-2.1-forward-port:
```

williamdalessandro (changes requested):

.github/forward-port-config.yml
aimeeu: OH - I totally did not get that from reading the forward port README.

williamdalessandro (approved):
LGTM!

PR #3515, opened 2026-09-22
VAULT-50237 docs(rar): caution against dangerous wildcard paths

jasonpilz:
## What

Adds caution language to the RAR (Rich Authorization Requests) docs clarifying that wildcard (`*`) support in the `vault:path_access` `path` field mirrors Vault ACL glob syntax for backward compatibility, and that it is the customer's Authorization Server / IdP / STS responsibility to keep wildcard scopes reasonable rather than dangerously broad when minting RAR tokens.

## Why

[VAULT-50237](https://hashicorp.atlassian.net/browse/VAULT-50237): RAR supports wildcards for parity with existing ACL syntax and to avoid a breaking change to released GA behavior, but docs didn't previously call out the risk of overly-broad wildcard grants or clarify whose responsibility it is to scope them safely.

## Changes

- `content/vault/v2.x/content/ai/iam/concepts/rar/index.mdx`: note that ACL intersection doesn't make a wildcard safe on its own.
- `content/vault/v2.x/content/ai/iam/concepts/rar/enforcement.mdx`: caution near the existing wildcard example.
- `content/vault/v2.x/content/ai/iam/reference/rar-type-specification.mdx`: corrects an inaccurate "exact path matching" claim and adds a new Wildcard paths section, distinguished from the existing templated-path wildcard rejection behavior.

Docs-only change; no behavior changes.

[VAULT-50237]: https://hashicorp.atlassian.net/browse/VAULT-50237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


content/vault/v2.x/content/ai/iam/concepts/rar/enforcement.mdx
```diff
@@ -97,6 +97,14 @@ For example, the following JWT allows:
 - Creating or updating at `pki/issue/my-role` (with `common_name` required)
 - Reading or listing from `secret/data/app-*` paths (except version 1)
 
+<Warning>
+
+The `secret/data/app-*` constraint below uses a wildcard path. Scope wildcard
+grants to a specific prefix rather than an entire mount. Refer to
```
VioletHynes, line 103: I don't know that this warning is super relevant here, since we are giving a prefix and not an entire mount

content/vault/v2.x/content/ai/iam/reference/rar-type-specification.mdx
```diff
@@ -33,6 +33,29 @@ policy-level controls, enabling fine-grained authorization directly within OAuth
 tokens.
 
 
+## Wildcard paths
+
+The `path` field supports wildcard (`*`) matching using the same glob syntax
+as Vault ACL policy paths. Vault enforces whatever scope a token contains and
+cannot distinguish a narrowly-scoped wildcard (for example,
```
VioletHynes, line 40: I think this wording is fairly weak --- Vault could if it wanted to distinguish a more narrowly scoped wildcard, but Vault won't ever know what task you need to do and whether or not the permissions are appropriately scoped. I think let's focus on the latter rather than the former

VioletHynes (approved):
Looks good! I suggested some small changes but they're mostly nits.

content/vault/v2.x/content/ai/iam/concepts/rar/enforcement.mdx
jasonpilz: Ok I've removed that now.

content/vault/v2.x/content/ai/iam/reference/rar-type-specification.mdx
jasonpilz: Have re-worded this, hopefully it's better now 👍🏻

VioletHynes (approved):
Looks good! Thanks for updating this.

jonathanfrappier (approved):

